PR #5, opened 2022-01-28
Feature/add post filters endpoint

ChrysmOre:
### What

- Added first endpoint POST /filters
- Added generator and responder packages
- Added datastore/mongodb and model packages
- Added middleware & authorization, public & private endpoints
- Added request & response contracts to api package, along with parseRequest helper including request validation check
- Added component tests, but not checking contents of document saved in Mongo (for future ticket)
- Added generic etag generation function and hasher interface (not actually used for POST endpoint)

### How to review

Check code makes sense and test pass. To experiment with API:
- Clone repo
- Can use mocked request body in `api/mock` to probe `/filters` endpoint with
`% cd api/mock`
`% curl -v -d @create_filter.request.json -H "AUTHORIZATION: $SERVICE_AUTH_TOKEN" "http://localhost:27100/filters" | jq`
 

### Who can review

Anyone


DavidSubiros (commented):
- Code looks very good
- Unit/component tests pass
- Tested with dp-compose, I could create a new filter and see it in mongo.

Things to improve:
- I tested the `/health` endpoint, and it requires auth, but it shouldn't. Could you please add an 'exception list' for auth middleware, please? Or maybe apply the auth middleware only on `/fiters` prefix?
- Should there be unit tests for the api package?

ChrysmOre:
> * Code looks very good
> * Unit/component tests pass
> * Tested with dp-compose, I could create a new filter and see it in mongo.
> 
> Things to improve:
> 
> * I tested the `/health` endpoint, and it requires auth, but it shouldn't. Could you please add an 'exception list' for auth middleware, please? Or maybe apply the auth middleware only on `/fiters` prefix?
> * Should there be unit tests for the api package?

Ah yeah I should be able to fix the health endpoint issue without too much trouble. And yes maybe some unit tests for the validation and request parsing would be good.

DavidSubiros:
One more observation:
If we don't need Kafka, then we can also remove the zookeeper container, which is only used by kafka nodes .

DavidSubiros (approved):
LGTM:
- healthcheck issue has been fixed
- component tests cover the api quite well, and there are unit tests for responder, etc.

DavidSubiros (approved):
LGTM

PR #8, opened 2022-02-17
Get filter dimensions

PauloPortugal:
### What
* Create endpoint in dp-flex-filter-api and utilise existing packages
* Implement mongodb
* Create data models (main filter model already exists)
* Implement endpoint functionality (retrieves dimensions and options information for a specific filter id)

### How to review

So far no tests have been added. will do this next.

### Who can review

Team B developers


api/filters.go
```diff
@@ -2,6 +2,7 @@ package api
 
 import (
 	"context"
+	"github.com/go-chi/chi/v5"
```
ChrysmOre, line 5: It's generally preferred to have some sorting with the imports along the lines of, - keeping standard library, ONS packages and third party packages in 2 separate groups and sorted alphabetically
```suggestion
import (
        "context"
        "net/http"
        "strconv"
        
	"github.com/go-chi/chi/v5"
```

api/filters.go
```diff
@@ -129,6 +130,26 @@ func (api *API) createFilter(w http.ResponseWriter, r *http.Request) {
...
+	ctx := r.Context()
+	fID := chi.URLParam(r, "id")
+
+	dimensions, err := api.store.GetFilterDimensions(ctx, fID)
+	if err != nil {
+		api.respond.Error(
+			ctx,
+			w,
+			statusCode(err),
+			errors.Wrap(err, "failed to get filter"),
```
ChrysmOre, line 143: Message should probably be "failed to get filter dimensions".

datastore/mongodb/filters.go
```diff
@@ -40,3 +40,18 @@ func (c *Client) CreateFilter(ctx context.Context, f *model.Filter) error {
...
+// GetFilterDimensions gets the dimensions for a Filter in the CantabularFilters collection
+func (c *Client) GetFilterDimensions(ctx context.Context, fID string) ([]model.Dimension, error) {
+	var err error
+
+	col := c.collections.filters
+
+	var f model.Filter
+
+	if err = c.conn.Collection(col.name).FindOne(ctx, bson.M{"id": fID}, f); err != nil {
+		return []model.Dimension{}, errors.Wrap(err, "failed to get filter")
```
ChrysmOre, line 53: You can return `nil` here rather than `[]model.Dimension{}`.

ChrysmOre: I think our dp-mongodb has an error type we can check against to determine if it's NotFound error `https://github.com/ONSdigital/dp-mongodb/blob/main/mongodb/error.go`. Similar to what's done in collections.go lock() function you can add some extra detail to the error here by using the package's error tpye and doing something like:
```
	if err != nil {
		err := &er{
			err: errors.Wrap(err, "failed to get filter"),
		}
		if errors.Is(err, dpmongo.ErrNoDocumentFound) {
			err.notFound = true
		}
        }
```
This will make the statusCode assert to 404 in the handler

api/filters.go
ChrysmOre: Consider whether we want the details of the error returned to the user. AS these endpoints will be facing the public exposing details of our database implemetation could open us up to people trying to find exploits. The api package's error type has the ability to return a different message to the user from the error that gets logged like:
```
return Error{
	err: errors.Wrap(err, "failed to get filter dimensions"), // gets logged for our purposes
	message: "failed to get filter dimensions", // singular message gets returned to user, no wrapped error details etc
	logData: log.Data{
	 "id": fID, // example, use log.Data to annotate errors with useful details for our logging
	}
}
```

PauloPortugal: Good point.
I have updated my IDE.
Resolved.

PauloPortugal: Good point as well.

api/error.go
```diff
@@ -45,14 +45,14 @@ func (e Error) LogData() map[string]interface{} {
...
 	switch {
-	case errors.As(err, &serr):
-		return serr.Code()
 	case dperrors.Unavailable(err):
 		return http.StatusServiceUnavailable
 	case dperrors.NotFound(err):
 		return http.StatusNotFound
 	case dperrors.Conflict(err):
 		return http.StatusConflict
+	case errors.As(err, &serr):
```
ChrysmOre, line 54: Why did you move this? This was purposely first so it would be evaluated first and take precedence

features/steps/steps.go
```diff
@@ -55,6 +62,25 @@ func (c *Component) theDocumentInTheDatabaseShouldBe(id string, doc *godog.DocSt
...
 }
 
+func (c *Component) iHaveTheseFilters(docs *godog.DocString) error {
+	ctx := context.Background()
+	var filters []model.Filter
+	m := c.store
+
+	err := json.Unmarshal([]byte(docs.Content), &filters)
+	if err != nil {
+		return err
```
ChrysmOre, line 72: Could you wrap these errors with some context? `errors.Wrap(err, "failed to blah")` etc. Thanks

PauloPortugal: this has been addressed.

api/error.go
PauloPortugal: with the previous order the error return was always a 500 and never a 404, based on the error code returned by the DAO/MongoDB pakage.

PauloPortugal: This has been addressed.

ChrysmOre (approved):
LGTM

PR #9, opened 2022-02-17
Feature/add type to filter struct and publishing checks

ChrysmOre:
### What

- Added `type` field to filter struct
- Added check to make sure only published datasets are available to create filters on for unauthenticated users

### How to review

Check changes make sense, test pass.

To test manually (optional):
- Spin up cantabular journey services in `dp-compose/cantabular-import-journey` with the environment variables `ENABLE_PRIVATE_ENDPOINTS` and `ENABLE_PERMISSIONS_AUTH` set to `false` for `dp-cantabular-filter-flex-api` (set in `dp-cantabular-filter-flex-api.yml`)
- Import a Cantabular Flexible tabe dataset but don't publish it
- Make a request to create a new filter without authentication:`cd api/mock` edit `filter.request.json` to point to you newly imported dataset, `curl -v -d @create_filter.request.json   http://localhost:27100/filters | jq`
- Inspect response for error response `dataset not found` and http status code 404. Repeat request with authentication and you should receive 201 success response (add `-H "Authorization: $SERVICE_AUTH_TOKEN"` to curl request)
 
### Who can review

- Anyone
- Anyone familiar with the Cantabular import journey for manual testing

### DIFF
--- a/api/filters.go
+++ b/api/filters.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/ONSdigital/dp-api-clients-go/v2/cantabular"
 	"github.com/ONSdigital/dp-api-clients-go/v2/dataset"
+	dprequest "github.com/ONSdigital/dp-net/v2/request"
 	"github.com/ONSdigital/log.go/v2/log"
 
 	"github.com/google/uuid"
@@ -17,6 +18,7 @@ import (
 
 const (
 	flexible = "flexible"
+	published = "published"
 )
 
 func (api *API) createFilter(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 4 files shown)


api/filters.go
```diff
@@ -52,6 +58,19 @@ func (api *API) createFilter(w http.ResponseWriter, r *http.Request) {
...
 	}
 
+	if v.State != published && !dprequest.IsCallerPresent(ctx) {
+		api.respond.Error(
+			ctx,
+			w,
+			http.StatusNotFound,
+			Error{
+				err:     errors.New("unauthenticated request on unpublished dataset"),
+				message: "dataset not found",
```
cookel2, line 68: The message seems a little misleading. Perhaps something like this would prevent confusion:
```suggestion
				message: "dataset only available to authorised users",
```

cookel2 (commented):
please see suggestion

api/filters.go
ChrysmOre: That was on purpose, I was thinking the public shouldn't be aware of unplublished datasets? The full error message with details regarding authentication does get logged for our purposes though (`err` gets logged, `message` is returned to caller)

cookel2: Yes I realise that. I just think it would be better not to give the caller a false impression. Fair enough that they shouldn't have access to the dataset. But being aware that it exists but is not published should be ok. So perhaps something like "dataset not published" might be better?

cookel2 (approved):

PR #12, opened 2022-03-01
add basic GET /filters/{id} endpoint

ChrysmOre:
### What

- Add GET /filters endpoint.

- Added missing health checks to service

- Some small updates to existing code & data models

### How to review

Check changes make sense, tests pass

### Who can review

Anyone


api/interface.go
```diff
@@ -30,9 +31,8 @@ type validator interface {
 }
 
 type generator interface {
-	PSK() ([]byte, error)
-	UUID() (uuid.UUID, error)
 	Timestamp() time.Time
+	UniqueTimestamp() primitive.Timestamp
```
PauloPortugal, line 35: Why do we need it?

features/mock/generator.go
```diff
@@ -34,6 +35,15 @@ func (g *Generator) Timestamp() time.Time {
...
 
+// Timestamp generates a constant timestamp
+func (g *Generator) UniqueTimestamp() primitive.Timestamp {
+	t, _ := time.Parse(time.RFC3339, TestTimestamp)
+	return primitive.Timestamp{
+		T: uint32(t.Unix()),
+		I: 1,
+	}
+}
+
```
PauloPortugal, line 46: Why do we need this?

api/interface.go
ChrysmOre: In the existing filters service/document this is how it's saved, there was an issue determining the filterType in the other service because this field it was unable to unmarshal a time that had been saved as a UTC time.time.

PR #15, opened 2022-03-03
Feature/api filter outputs

mohsdev:
### What

Describe what you have changed and why.

### How to review

Describe the steps required to test the changes.

### Who can review


api/filter_outputs.go
```diff
@@ -0,0 +1,28 @@
...
+		api.respond.Error(ctx, w, http.StatusBadRequest, errors.Wrap(err, "parse request"))
+		return
+	}
+	if err := api.validate.Struct(&request); err != nil {
+		api.respond.Error(ctx, w, http.StatusBadRequest, errors.Wrap(err, "validate request fields"))
+		return
+	}
+
+	if err := api.store.CreateFilterOutput(ctx, request.Downloads); err != nil {
+		api.respond.Error(ctx, w, http.StatusBadRequest, errors.Wrap(err, "upsert filter output"))
```
ChrysmOre, line 23: StatusBadRequest?

api/filter_outputs.go
```diff
@@ -0,0 +1,28 @@
...
+func (api *API) createFilterOutput(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+
+	var request createFilterOutputRequest
+	if err := api.ParseRequest(r.Body, &request); err != nil {
+		api.respond.Error(ctx, w, http.StatusBadRequest, errors.Wrap(err, "parse request"))
+		return
+	}
+	if err := api.validate.Struct(&request); err != nil {
+		api.respond.Error(ctx, w, http.StatusBadRequest, errors.Wrap(err, "validate request fields"))
```
ChrysmOre, line 18: The request objects are set up to have a valid() function which automatically gets called by parse request, it would be better to include any validation in that function

api/filter_outputs.go
ChrysmOre, line 23: Please can you make the error messages more human readable i.e `failed to create filter output` rather than just `upsert filter output`. Also we don't want to be returning database implementation details to the user in the error response, see the existing endpoints for how to tailor a separate message for the caller to what gets logged. You might want to think about adding log data to your errors as well.

datastore/mongodb/filter_outputs.go
```diff
@@ -0,0 +1,39 @@
...
+
+	col := c.collections.filters
+	lockID, err := col.lock(ctx, filterOutput.ID.String())
+	if err != nil {
+		return err
+	}
+	defer col.unlock(ctx, lockID)
+
+	if _, err := c.FilterOutputs().UpsertById(ctx, filterOutput.ID, bson.M{"$set": filterOutput}); err != nil {
+		errors.Wrap(err, "failed to upsert filter output")
```
ChrysmOre, line 36: You've forgotten `return` here.. You should test your code builds before submitting a PR

ChrysmOre (changes requested):

datastore/mongodb/filter_outputs.go
mohsdev: done

api/filter_outputs.go
mohsdev: done

mohsdev: Sorry, should be StatusInternalServerError

api/filter_outputs.go
mohsdev: hmmm, let me check

mohsdev: done

mohsdev:
@ChrysmOre already fix the review suggestions, help to review

PR #18, opened 2022-03-03
Added support for filter-outputs POST endpoint

j-s-rawat:
What

https://trello.com/c/yC28daUV/5468-create-post-filter-outputs-endpoint

How to review
The changes support POST endpoint for 'filter-outputs'
A UT added for validation, and currently the only validation check is to make sure that input body has non-empty data

Who can review
Any back-end dev that has previously worked on endpoints


api/contract.go
```diff
@@ -44,6 +45,59 @@ type createFilterResponse struct {
 	model.Filter
 }
 
+// createFilterRequest is the request body for POST /filters
+type createFilterOutputsRequest struct {
+	State     string             `bson:"state" json:"state"`
+	Downloads model.FilterOutput `bson:"downloads"      json:"downloads"`
+}
+
+func isNotFullyPopulated(fi *model.FileInfo) error {
```
ChrysmOre, line 54: This function would be better as a method on the *model.FileInfo type, as it has quite a generic name but takes a specific type

datastore/mongodb/filter_outputs.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"context"
+
+	"github.com/ONSdigital/dp-cantabular-filter-flex-api/model"
+	"github.com/pkg/errors"
+	"go.mongodb.org/mongo-driver/bson"
+)
+
+// CreateFilterOutput creates a new FilterOutputs in the CantabularFilters colllection
+func (c *Client) CreateFilterOutput(ctx context.Context, f *model.FilterOutputResponse) error {
```
ChrysmOre, line 12: Ah I see you've create a FilterOutputResponse struct in package model and used it here as a database model.. The FilterOutputResponse type should be in package api contract.go, and should be only used for responding to the request. The database model should be in package model as you have it but it shouldn't be called FilterOutputResponse, it should be called FilterOutput. If the fields are identical you can embed FilterOutput in FilterOutputResponse as is done with createFilterRequest.
Also it should be `getFilterOutputResponse`.

ChrysmOre (changes requested):
Some comments

datastore/mongodb/filter_outputs.go
```diff
@@ -0,0 +1,28 @@
...
+
+	if id, err := c.generate.UUID(); err != nil {
+		return errors.Wrap(err, "failed to generate UUID: %w")
+	} else {
```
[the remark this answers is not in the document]
ChrysmOre: Probably better to not initialse with 'if'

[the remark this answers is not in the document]
j-s-rawat: Use else as local variable 'id' is declared inline in the 'if' clause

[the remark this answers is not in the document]
j-s-rawat: Will change to match the other way of declaration

api/contract.go
```diff
@@ -49,6 +49,34 @@ type getFilterResponse struct {
 	model.Filter
 }
 
+// createFilterResponse is the response body for POST /filters
+type createFilterOutputResponse struct {
+	model.FilterOutput
+}
+
+// createFilterRequest is the request body for POST /filters
+type createFilterOutputsRequest struct {
```
ChrysmOre, line 58: Should be `createFilterOutputRequest` singular, and comment is incorrect. JSON tags could be lined up too.

api/contract.go
```diff
@@ -49,6 +49,34 @@ type getFilterResponse struct {
 	model.Filter
 }
 
+// createFilterResponse is the response body for POST /filters
```
ChrysmOre, line 52: comment is incorrect

api/filter_outputs.go
```diff
@@ -0,0 +1,59 @@
...
+
+	"github.com/ONSdigital/dp-cantabular-filter-flex-api/model"
+	dprequest "github.com/ONSdigital/dp-net/v2/request"
+	"github.com/pkg/errors"
+)
+
+func (api *API) createFilterOutput(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+
+	if !dprequest.IsCallerPresent(ctx) {
```
ChrysmOre, line 14: Sorry to mess around, I think this is actually not needed. Instead it we can just remove the public endpoint entirely. You will need to update your unauthorized test slightly (or maybe even remove it)

ChrysmOre (commented):
Looking good, just a few more things

api/contract_test.go
```diff
@@ -84,3 +84,61 @@ func TestCreateFiltersRequestValid(t *testing.T) {
 		})
 	})
 }
+
+func TestCreateFilterOutputsRequestValid(t *testing.T) {
+
+	Convey("Given a valid createFilterOutputsRequest request object", t, func() {
```
goofballLogic, line 90: this doesn't fully test the code under test. Is that ok?

api/filter_outputs.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/ONSdigital/dp-cantabular-filter-flex-api/model"
+	"github.com/pkg/errors"
+)
+
+func (api *API) createFilterOutput(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+
+	var req createFilterOutputRequest
+
+	if err := api.ParseRequest(r.Body, &req); err != nil {
```
goofballLogic, line 15: Some of the errors encountered by ParseRequest may not be BadRequest ones but we always return that status.

datastore/mongodb/filter_outputs.go
```diff
@@ -0,0 +1,27 @@
...
+	id, err := c.generate.UUID()
+	if err != nil {
+		return errors.Wrap(err, "failed to generate UUID: %w")
+	}
+
+	f.ID = id.String()
+
+	col := c.collections.filterOutputs
+
+	if _, err := c.conn.Collection(col.name).UpsertById(ctx, f.ID, bson.M{"$set": f}); err != nil {
```
goofballLogic, line 22: We need to check that the item was created here. If it was not (e.g. something got updated), that is an error

api/contract_test.go
j-s-rawat: We are just trying to test simple validation function, which is currently testing if any of the field is blank. Rest of the functionality is being tested in functional test

features/filter_outputs_authorised.feature
```diff
@@ -0,0 +1,172 @@
...
+    {
+      "errors": [
+        "badly formed request body: unexpected end of JSON input"
+      ]
+    }
+    """
+
+    And the HTTP status code should be "400"
+    
+Scenario: Creating a new filter empty request body
```
goofballLogic, line 117: this isn't an empty request body

features/filter_outputs_authorised.feature
```diff
@@ -0,0 +1,172 @@
...
+    Then I should receive the following JSON response:
+    """
+    {
+      "errors": [
+        "failed to parse request: invalid request: \"Public\" is empty in input"
+      ]
+    }
+    """
+
+    And the HTTP status code should be "400"
```
goofballLogic, line 172: no test for failing to create the filter

api/filter_outputs.go
j-s-rawat: The chances of request failing with io.read or json.Unmarshal, for a reason not related to bad input body is very slim.  As errors.Wrap is presenting the full stack trace, it will pinpoint the cause of error so a single status for every situation won't be bad.

goofballLogic: I disagree but happy to accept your conclusion

datastore/mongodb/filter_outputs.go
goofballLogic: I disagree with not inspecting the output but happy to accept your conclusion

goofballLogic (approved):
I've agreed to disagree with @j-s-rawat on comments. Test coverage now appears to be pretty good so approving.

api/filter_outputs_test.go
```diff
@@ -0,0 +1,126 @@
...
+		  "private" : "http://minio:9000/private-bucket/datasets/cantabular-flexible-example-2021-1.txt",
+		  "public" : "https://csv-exported.s3.eu-west-1.amazonaws.com/full-datasets/cpih01-time-series-v1.csv-metadata.txt",
+		  "size" : "530",
+		  "skipped": true
+		}
+	}
+}`)
+
+type FuncReturnErrorAndCode func() (int, error)
+
```
ChrysmOre, line 55: Wouldn't these mocks go in the `mock` package?

ChrysmOre (changes requested):
Looks as if previous comments are addressed but have one more

ChrysmOre (approved):

PR #20, opened 2022-03-04
Updates to swagger spec for future tickets

franmoore05:
### What

Updated the model for filter-outputs, added NewFilterResponse, and updated POST filter-outputs to be PUT to bring things in to line with existing dp-filter-api.

Added GET, PUT, PATCH, DELETE dimensions as per dp-filter-api.

Also updated the PUT filter endpoint to reflect latest standards.

### How to review

Ensure the changes make sense.

### Who can review

Anyone.

### DIFF
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -234,34 +234,6 @@ paths:
           $ref: '#/responses/FilterNotFound'
         '500':
           $ref: '#/responses/InternalError'
-    put:
-      tags:
-      - "Public"
-      summary: "Update a filter"
-      description: "Update the filter by providing new properties, submit a filter for processing by setting query parameter `submitted` to `true`"
-      parameters:
-      - $ref: '#/parameters/submitted'
-      - $ref: '#/parameters/update_filter'
-      - $ref: '#/parameters/if_match'
-      responses:
-        200:
-          description: "The filter job has been updated"
-          schema:
-            $ref: '#/definitions/UpdateFilterResponse'
-          headers:
-            ETag:
-              type: string
-              description: "Defines a unique filter resource version"
-        400:
-          description: "Invalid request body or If-Match header not provided"
-        404:
-          $ref: '#/responses/FilterNotFound'
-        409:
-          description: '#/responses/FilterConflict'
-        422:
-          description: "Unprocessable entity - instance has been removed"
-        500:
-          $ref: '#/responses/InternalError'
   /filters/{id}/dimensions:
     get:
       tags:


ChrysmOre (changes requested):
Just going to leave comments for each endpoint here.
- POST /Filters:
  - I don't think dimension_url is required in the POST body, it should be generated/got from version doc (as per dp-filter-api).
  - We don't want to use 'cantabular_blob' for any field names that face users, it needs to be 'population_type'
- GET /filters/{id}
  - Seems to be missing `dimension_list_url`
- PUT /filters/{id}  
    - Seems to be missing `dimension_list_url`
- GET /filters/{id}/dimensions/
  - Existing endpoint only seems to have name and url to individual dimension rather than include options and extra info. Although is this because existing filter journey wouldn't save list of options but we do need to?
 - POST /query
   - `cantabular_blob` needs to be `population_type` 
   - does 'dimension_options' need to be 'dimensions'?
   
Aaaand new line missing at end of file :)

franmoore05:
> Just going to leave comments for each endpoint here.
> 
> * POST /Filters:
>   
>   * I don't think dimension_url is required in the POST body, it should be generated/got from version doc (as per dp-filter-api).
>   * We don't want to use 'cantabular_blob' for any field names that face users, it needs to be 'population_type'
> * GET /filters/{id}
>   
>   * Seems to be missing `dimension_list_url`
> * PUT /filters/{id}
>   
>   * Seems to be missing `dimension_list_url`
> * GET /filters/{id}/dimensions/
>   
>   * Existing endpoint only seems to have name and url to individual dimension rather than include options and extra info. Although is this because existing filter journey wouldn't save list of options but we do need to?
> * POST /query
>   
>   * `cantabular_blob` needs to be `population_type`
>   * does 'dimension_options' need to be 'dimensions'?
> 
> Aaaand new line missing at end of file :)

I've updated the file with the above comments - although for the POST /query endpoint will we not need the dimension_options for when the user changes coverage?

goofballLogic (changes requested):
The "dimensions" here is wrapped in a paged "items" list but needs to match filter-api's one, like this:

```
"dimensions": [
    {
      "name": "string",
      "options": [
        "string"
      ]
    }
  ]
```

ChrysmOre (dismissed):

swagger.yaml
```diff
@@ -218,34 +236,31 @@ paths:
...
       tags:
-        - Public
-        - Private
-      summary: Update a filter
-      description: >-
-        Update the filter by providing new properties, submit a filter for
-        processing by setting query parameter `submitted` to `true`
+      - "Public"
+      summary: "Update a filter"
+      description: "Update the filter by providing new properties, submit a filter for processing by setting query parameter `submitted` to `true`"
```
ChrysmOre, line 241: This process isn't true for us. This was a hack in the old api Eleanor asked my to change if I could (I didn't in the existing one but no point copying it in ours), we're going to have the separate /query endpoint to trigger ours. I don't think we need any of those URL parameters. No idea what that if-match parameter is all about it's actually a header.

franmoore05: This has now been removed.

ChrysmOre (approved):

franmoore05:
> "dimensions": [
>     {
>       "name": "string",
>       "options": [
>         "string"
>       ]
>     }
>   ]

This should now have been amended.

goofballLogic (approved):
Was approved by another team member

PR #21, opened 2022-03-09
Add pagination to dimensions in Filter Flex API (Ticket #5559)

jayfreestone:
### What

Updates the dimensions endpoint to conform to the ONS pagination standards.

Since dimensions are nested inside filters, we used `unwind` and `facet` to pull out the relevant info. Unlike `findOne`, aggregates always seem to return a list of results, and therefore don't seem to provide an easy way for us to determine if the top-level filter is missing. We've added a flag which gets set in the event that any matches are returned, but another alternative would be running a separate `Count()` query first.

(First time working with Mongo, so any alternatives/suggestions welcome.)

This is effectively a breaking change, since it renames `dimensions` to `items`. However, the client is [already set up](https://github.com/ONSdigital/dp-api-clients-go/blob/2b9ec3cde38ebe3b0e37971bff49cf154969f566/filter/data.go#L14) to look for items (and therefore never returns any results), so it should be safe to change.

Resolves: [5559](https://trello.com/c/vsEwaNIe)

### How to review

Create a filter on a dataset, then paginate through the results by changing the `offset`/`limit` query params.

Example: `http://localhost:27100/filters/4661c9a7-d24d-4c30-be71-8aa6cb72b9e9/dimensions?limit=1&offset=0`

```json
{
    "items": [
        {
            "name": "City",
            "options": [
                "0"
            ],
            "dimension_url": "http://localhost:22000/datasets/cantabular-flexible-example/editions/2021/versions/1/dimensions/city/options",
            "is_area_type": true
        }
    ],
    "limit": 1,
    "offset": 0,
    "count": 1,
    "total_count": 2
}
```

### Who can review

Anyone.

### DIFF
--- a/api/contract.go
+++ b/api/contract.go
@@ -78,5 +78,14 @@ func (r *createFilterOutputRequest) Valid() error {
 
 // getFilterDimensionsResponse is the response body for GET /filters/{id}/dimensions
 type getFilterDimensionsResponse struct {
-	Dimensions []model.Dimension `json:"dimensions"`
+	Items []model.Dimension `json:"items"`
+	paginationResponse
+}
+
+// paginationResponse represents pagination data as returned to the client.
+type paginationResponse struct {
+	Limit      int `json:"limit"`
+	Offset     int `json:"offset"`
+	Count      int `json:"count"`
+	TotalCount int `json:"total_count"`
 }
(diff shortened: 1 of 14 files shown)


PauloPortugal:
@jayfreestone I really like your commit hygiene. We should use the `Resolves` tag :+1:

PauloPortugal (approved):
Well implemented and documented.

PauloPortugal (approved):

PR #24, opened 2022-03-11
Update Filter Flex API dimensions pagination to avoid $facet (Ticket #5559)

jayfreestone:
### What

`$facet` [isn't supported](https://docs.aws.amazon.com/documentdb/latest/developerguide/mongo-apis.html) by DocumentDB, meaning the latest [pagination changes](https://github.com/ONSdigital/dp-cantabular-filter-flex-api/pull/21) don't work on develop.

We didn't notice the error since locally we use Mongo (either in a container, or in-memory for testing). There doesn't seem to be a way to run DocumentDB locally, so not sure how this could be avoided other than carefully checking the compatibility list every time you make a change 😕.

### How to review

This PR just replicates the behaviour from the [previous PR](https://github.com/ONSdigital/dp-cantabular-filter-flex-api/pull/21), but without using `$facet`, so it can be tested the same way.

### Who can review

Anyone.


franmoore05 (approved):
LGTM

PR #25, opened 2022-03-14
Add filter dimension POST endpoint

PauloPortugal:
### What

 * Fix Swagger documentation

### How to review

Review swagger

### Who can review

Anyone


goofballLogic:
> 

@jayfreestone we're doing a POST to a collection of dimensions. It wouldn't be right to identify the URL of the resource being created would it?

jayfreestone (approved):
Nice one, LGTM 👍

One super minor nitpick: is it allowed to create a filter with 0 options? If so, I imagine that the following kind of payload:

```json
  {
    "name": "Number of siblings (3 mappings)"
  }
```

Will result in this successful response:

```json
  {
    "name": "Number of siblings (3 mappings)",
    "is_area_type": false,
    "options": null
  }
```

Might be worth ensuring `options` is always an empty array when it deserialises (or that the store dimension is always an empty array).

PauloPortugal:
@jayfreestone I have addressed your concern.
Please review again.

jayfreestone (approved):

andre-urbani (approved):

PR #26, opened 2022-03-15
fix bug shutting down bdd component

goofballLogic:
### What

Ensure that we don't send the interrupt signal to the service twice (as it's a blocking message and will hang)

### How to review

Read, think

### Who can review

gophers other than @goofballLogic

### DIFF
--- a/features/steps/component.go
+++ b/features/steps/component.go
@@ -146,7 +146,6 @@ func (c *Component) startService(ctx context.Context) {
 		}()
 		panic(fmt.Errorf("unexpected error received from errorChan: %w", err))
 	case sig := <-c.signals:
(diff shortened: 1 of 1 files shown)


PauloPortugal (commented):
Can you `git fixup` you **oops** commit message, otherwise we will have a oops  commit message on the repo, when the previous commit could have been amended.

```
git rebase -i HEAD~2
```
then replace the second commit `pick` with `f`

you will have to force push to the remote repo.

goofballLogic:
I'm happy for my oops commit to be in the repo. This is part of the coding journey.

PauloPortugal (approved):

PR #27, opened 2022-03-16
Adding PUT filters endpoint

franmoore05:
### What

Added a new PUT filters/{id} endpoint as it was missing from the original spec.

### How to review

Ensure the changes make sense.

### Who can review

Anyone.


andre-urbani (approved):
LGTM

PR #29, opened 2022-03-17
Enable POST '/filter/{id}/dimensions' private endpoint

PauloPortugal:
### What
We should enable the private endpoint for POST '/filter/{id}/dimensions'
Resolves: [5568](https://trello.com/c/6L3QBhtd/5568-create-post-filters-id-dimensions-name-endpoint)
### How to review

 * Runs tests 

### Who can review

any developer


franmoore05 (approved):
LGTM

PR #40, opened 2022-03-29
add dataset endpoint

s-xton:
### What

Added sample flex datasets endpoint

### How to review

Please review the sample endpoint against card #5483 Create new /json endpoint in dp-filter-flex-api

### Who can review

Anyone


franmoore05 (approved):
LGTM

PR #42, opened 2022-03-29
Added endpoint /filter-outputs/{filter_output_id}/events

j-s-rawat:
### What
Story: https://trello.com/c/rST9QPho/5605-create-post-filter-outputs-filter-output-id-events
New endpoint for adding/appending the events in the existing filter-outputs record
### How to review
Check that the endpoint meets the swagger spec specification
There is a functional test that covers all the possible response in the swagger specs

### Who can review

Anyone


api/contract.go
```diff
@@ -75,6 +75,10 @@ type createFilterOutputRequest struct {
 	model.FilterOutput
 }
 
+type eventRequest struct {
```
ChrysmOre, line 78: Please specify create/delete/get etc for contract structs

model/filter.go
```diff
@@ -56,9 +56,16 @@ type Link struct {
...
-	Downloads Downloads `bson:"downloads,omitempty"   json:"downloads,omitempty"`
+	ID         string            `bson:"id"                      json:"id"`
+	FilterID   string            `bson:"filter_id"               json:"filter_id"`
+	InstanceID string            `bson:"instance_id"             json:"instance_id"`
+	Dataset    Dataset           `bson:"dataset,omitempty"       json:"dataset,omitempty"`
+	Dimensions []Dimension       `bson:"dimensions,omitempty"    json:"dimensions,omitempty"`
+	Downloads  Downloads         `bson:"downloads,omitempty"     json:"downloads,omitempty"`
+	Events     []Event           `bson:"events,omitempty"        json:"events,omitempty"`
+	Links      FilterOutputLinks `bson:"links,omitempty"         json:"links,omitempty"`
+	Published  bool              `bson:"published,omitempty"     json:"published,omitempty"`
```
ChrysmOre, line 67: will need to be a pointer to bool for omitempty to work. Also Dataset, Downloads and LInks will need to be pointers too for same reason

api/api.go
```diff
@@ -71,6 +71,7 @@ func (api *API) enablePrivateEndpoints() {
 	r.Put("/filters/{id}", api.putFilter)
 	r.Get("/filters/{id}/dimensions", api.getFilterDimensions)
 	r.Post("/filters/{id}/dimensions", api.addFilterDimension)
-	r.Put("/filter-outputs/{filter-output-id}", api.updateFilterOutput)
+	r.Put("/filter-outputs/{filter_output_id}", api.updateFilterOutput)
```
j-s-rawat, line 74: Not related for this story but from previous changes, dash to underscore

features/put_filter_outputs.feature_put_filter_outputs_needs_fixing
```diff
@@ -49,9 +47,7 @@ Feature: Put Filter Outputs Private Endpoints Enabled
       },
       {
         "id": "94310d8d-72d6-492a-bc30-27584627edb2",
-        "filter_id": "94310d8d-72d6-492a-bc30-27584627fil2",
-        "instance_id": "94310d8d-72d6-492a-bc30-27584627inst2",
-        "state": "completed",
+        "state": "string",
```
PauloPortugal, line 50: It would be good to use a more real **state**, such as `completed` or `published`

features/put_filter_outputs.feature_put_filter_outputs_needs_fixing
```diff
@@ -91,7 +87,7 @@ Feature: Put Filter Outputs Private Endpoints Enabled
     ]
     """
 
-Scenario: PUT filter outputs successfully with existing ID in DB
+  Scenario: PUT filter outputs successfully with existing ID in DB
```
PauloPortugal, line 90: do we need this empty space?

PauloPortugal (commented):
Just a few comments, nothing major. Looks good to me.
Happy to approve you reply to the comments.

features/put_filter_outputs.feature_put_filter_outputs_needs_fixing
j-s-rawat: No not really

api/contract.go
j-s-rawat: done

api/filter_outputs.go
```diff
@@ -41,9 +41,49 @@ func (api *API) updateFilterOutput(w http.ResponseWriter, r *http.Request) {
...
 	}
 	api.respond.StatusCode(w, http.StatusOK)
 }
+
+//createFilterOutputEvent will add a new event to the list of existing events in the filter outputs
+func (api *API) addFilterOutputEvent(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	fID := chi.URLParam(r, "filter_output_id")
+
+	var req createEventRequest
```
ChrysmOre, line 56: Please keep the contract models precise to the handler names. `addFilterOutputEvent` -> `addFilterOutputEventRequest` etc

PauloPortugal (dismissed):

api/filter_outputs.go
```diff
@@ -41,9 +41,49 @@ func (api *API) updateFilterOutput(w http.ResponseWriter, r *http.Request) {
...
+				err: errors.Wrap(err, "invalid request body"),
+				logData: log.Data{
+					"id": fID,
+				},
+			},
+		)
+		return
+	}
+
+	f := model.Event{
```
ChrysmOre, line 73: As the request consists of an embedded `model.Event` you can just do `f := req.Event` here.
Also `f` isn't really an appropriate name, `f` was used for filter, `e` would probably be better as it's an event

datastore/mongodb/filter_outputs.go
```diff
@@ -31,47 +31,77 @@ func (c *Client) CreateFilterOutput(ctx context.Context, f *model.FilterOutput)
...
+		return err
+	}
+
+	return nil
+}
+
+//AddFilterOutputEvent will append to the existing list of events in the filter output
+func (c *Client) AddFilterOutputEvent(ctx context.Context, id string, f *model.Event) error {
+	col := c.collections.filterOutputs
+	docs, err := c.findFilterOutput(id, ctx)
```
ChrysmOre, line 73: Is it necessary to fetch some filter outputs to reinsert? Is there not away to update with one mongo query as we do elsewhere?

ChrysmOre (changes requested):
A few comments

datastore/mongodb/filter_outputs.go
```diff
@@ -31,47 +31,80 @@ func (c *Client) CreateFilterOutput(ctx context.Context, f *model.FilterOutput)
...
+	if err != nil {
+		return err
 	}
-
 	//a record with stat 'completed' can't be updated further
 	if docs[0].State == model.Completed {
-		return &er{
-			err:       errors.Errorf(`filter output is already in "completed" state`),
-			forbidden: true,
+		err := &er{
```
ChrysmOre, line 40: What was wrong with how this was done before?

datastore/mongodb/filter_outputs.go
```diff
@@ -31,47 +31,80 @@ func (c *Client) CreateFilterOutput(ctx context.Context, f *model.FilterOutput)
...
 }
+
+func (c *Client) findFilterOutput(fId string, ctx context.Context) ([]model.FilterOutput, error) {
+	col := c.collections.filterOutputs
+	var docs []model.FilterOutput
+	sc := bson.M{"id": fId}
+	var err error
+	var num int
+	if num, err = c.conn.Collection(col.name).Find(ctx, sc, &docs, mongodb.Limit(1)); err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
j-s-rawat: We don't have a specific error to message for this case, so can't see the benefit of wrapping this. 
The wrap will look like something like this 
&er{
err : err
}

datastore/mongodb/filter_outputs.go
j-s-rawat: It is just an outcome of bad merge

datastore/mongodb/filter_outputs.go
```diff
@@ -31,47 +31,80 @@ func (c *Client) CreateFilterOutput(ctx context.Context, f *model.FilterOutput)
...
-		return errors.Wrap(err, "failed to update filter output")
-	}
-	if num < 1 {
-		return &er{
-			err:      errors.Errorf("failed to find filter output"),
-			notFound: true,
-		}
+	docs, err := c.findFilterOutput(f.ID, ctx)
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
j-s-rawat: I am just propogating the error that is being returned. Don't think we need another layer of wrapper

datastore/mongodb/filter_outputs.go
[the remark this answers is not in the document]
ChrysmOre: I'm not sure what you mean by that? We want to wrap any error with a brief explanation of what's causing the error, this is so we understand precisely what the error is pointing to. Unless it will be redundant (i.e. no point returning `failed to find filter: failed to find filter`), we always want to add the extra information at each step so at the end we have complete description of what's gone wrong. Wrapping the error is about constructing the full error string, not just adding extra information like statusCodes and logData.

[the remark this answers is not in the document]
ChrysmOre: At the moment an error might read something like `failed add filter output event: invalid query` but that doesn't explain the exact database call that's failing as you're making more than one, if you wrap at each level it would read either `failed add filter output event: failed to get filter output: invalid query` or `failed add filter output event: failed to add element to filter output: invalid query`, so you can see it helps us to differentiate what's failed

datastore/mongodb/filter_outputs.go
```diff
@@ -75,3 +64,45 @@ func (c *Client) UpdateFilterOutput(ctx context.Context, f *model.FilterOutput)
...
+	col := c.collections.filterOutputs
+
+	ne := map[string]model.Event{
+		"events": *f,
+	}
+
+	var rec *mongodb.CollectionUpdateResult
+	var err error
+	if rec, err = c.conn.Collection(col.name).Update(ctx, bson.M{"id": id}, bson.M{"$push": ne}); err != nil {
+		return errors.Wrap(err, "failed to add element to filter output")
```
ChrysmOre, line 79: I think `failed to add event to filter output` would be more helpful?

features/put_filter_output_event.feature
```diff
@@ -0,0 +1,179 @@
...
+    When I POST "/filter-outputs/94310d8d-72d6-492a-bc30-27584627edb1/events"
+    """
+    {
+      "timestamp": "2016-07-17T08:38:25.319+0000",
+      "name": "cantabular-export-start"
+    }     
+    """
+
+    Then the HTTP status code should be "500"
+
```
ChrysmOre, line 179: Still a couple of stray spaces here, the file needs to end on a newline

ChrysmOre (approved):

andre-urbani (dismissed):

andre-urbani (approved):

PR #43, opened 2022-03-29
Fix/post filter unit tests

saminahbab:
### What

Fixed broken unit test for graceful service shutdown 
Added Kafka Producer Mock with Close function 

### How to review

run `make test` and `make test-component`

### Who can review

Team B

### DIFF
--- a/service/service_test.go
+++ b/service/service_test.go
@@ -232,33 +232,33 @@ func TestClose(t *testing.T) {
 			So(serverMock.ShutdownCalls(), ShouldHaveLength, 1)
 		})
 		/*
-		   TODO - figure out if this has been added to the boilerplate code or been removed from the csv exporter,
-		   such that, do we need to make this work here
+			   TODO - figure out if this has been added to the boilerplate code or been removed from the csv exporter,
+			   such that, do we need to make this work here
 
-		   		Convey("If service times out while shutting down, the Close operation fails with the expected error", func() {
-		   			cfg.GracefulShutdownTimeout = 1 * time.Millisecond
-		   			timeoutServerMock := &mock.HTTPServerMock{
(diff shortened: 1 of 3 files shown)


service/service_test.go
```diff
@@ -227,33 +232,33 @@ func TestClose(t *testing.T) {
...
+							Server:      timeoutServerMock,
+							HealthCheck: hcMock,
+						}
+
+						err = svc.Close(context.Background())
+						So(err, ShouldNotBeNil)
+						So(err.Error(), ShouldResemble, "context deadline exceeded")
+						So(len(hcMock.StopCalls()), ShouldEqual, 1)
+						So(len(timeoutServerMock.ShutdownCalls()), ShouldEqual, 1)
+					})*/
```
PauloPortugal, line 262: Please remove the added tab/space.
Do we need it?

PauloPortugal (commented):
It would be nice to include an explanation on the commit body, on why this failed and what was the fix for it.

This will help other developers understand a similar issue in other services.

service/service_test.go
saminahbab: Gofmt does get a bit confused on my machine tabbing commented code.
I would be happy to remove the commented block, as that is cleaner. Would you prefer that?

PauloPortugal: Yes I rather remove commented code.

saminahbab: Commented out code removed.

saminahbab:
@PauloPortugal https://github.com/ONSdigital/dp-cantabular-filter-flex-api/pull/43/commits/4de95d226641b0e727836e9cf45fa5ef54f87790 now has a commit message describing the reason for the fix.

PauloPortugal (approved):

PR #46, opened 2022-04-01
Add PUT dimensions output endpoint (Ticket #5623)

jayfreestone:
### What

Adds a new endpoint enabling the update of a dimension on a filter.

Resolves: [5623](https://trello.com/c/vUgHxQLq)

### How to review

Confirm the changes match the API spec.

### Who can review

Anyone.

### DIFF
--- a/api/error.go
+++ b/api/error.go
@@ -51,6 +51,8 @@ func statusCode(err error) int {
 		return http.StatusServiceUnavailable
 	case dperrors.NotFound(err):
 		return http.StatusNotFound
+	case dperrors.BadRequest(err):
+		return http.StatusBadRequest
 	case dperrors.Conflict(err):
 		return http.StatusConflict
 	case dperrors.Forbidden(err):
--- a/api/error.go
+++ b/api/error.go
@@ -9,9 +9,11 @@ import (
 
 // Error is the packages error type
 type Error struct {
-	err     error
-	message string
-	logData map[string]interface{}
+	err        error
+	message    string
+	logData    map[string]interface{}
+	notFound   bool
+	badRequest bool
 }
 
 // Error satisfies the standard library Go error interface
@@ -40,6 +42,16 @@ func (e Error) Message() string {
 // log data from an error
 func (e Error) LogData() map[string]interface{} {
 	return e.logData
+}
+
+// NotFound satisfies the errNotFound interface
+func (e Error) NotFound() bool {
+	return e.notFound
+}
+
+// BadRequest satisfies the errBadRequest interface
+func (e Error) BadRequest() bool {
+	return e.badRequest
 }
 
 func statusCode(err error) int {
--- a/api/filters.go
+++ b/api/filters.go
@@ -8,16 +8,15 @@ import (
 	"strings"
 	"time"
 
+	"github.com/ONSdigital/dp-api-clients-go/v2/cantabular"
+	"github.com/ONSdigital/dp-api-clients-go/v2/dataset"
 	dperrors "github.com/ONSdigital/dp-cantabular-filter-flex-api/errors"
 	"github.com/ONSdigital/dp-cantabular-filter-flex-api/event"
 	"github.com/ONSdigital/dp-cantabular-filter-flex-api/model"
-
-	"github.com/ONSdigital/dp-api-clients-go/v2/cantabular"
-	"github.com/ONSdigital/dp-api-clients-go/v2/dataset"
 	dprequest "github.com/ONSdigital/dp-net/v2/request"
 	"github.com/ONSdigital/log.go/v2/log"
-	"github.com/go-chi/chi/v5"
 
+	"github.com/go-chi/chi/v5"
 	"github.com/pkg/errors"
 )
 
@@ -412,6 +411,60 @@ func (api *API) addFilterDimension(w http.ResponseWriter, r *http.Request) {
 
 	w.Header().Set(eTagHeader, b)
 	api.respond.JSON(ctx, w, http.StatusCreated, resp)
+}
+
+func (api *API) updateFilterDimension(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	filterID := chi.URLParam(r, "id")
+	dimensionName := chi.URLParam(r, "name")
+
+	logData := log.Data{
+		"filter_id":      filterID,
+		"dimension_name": dimensionName,
+	}
+
+	updatedDimension := model.Dimension{
+		Options: []string{},
+	}
+
+	if err := api.ParseRequest(r.Body, &updatedDimension); err != nil {
+		api.respond.Error(
+			ctx,
+			w,
+			http.StatusBadRequest,
+			Error{
+				err:     errors.Wrap(err, "failed to parse update filter request"),
+				logData: logData,
+			},
+		)
+		return
+	}
+
+	var eTag string
+	if reqETag := api.getETag(r); reqETag != eTagAny {
+		eTag = reqETag
+	}
+
+	newETag, err := api.store.UpdateFilterDimension(ctx, filterID, dimensionName, updatedDimension, eTag)
+	if err != nil {
+		api.respond.Error(
+			ctx,
+			w,
+			statusCode(err),
+			Error{
+				err:     errors.Wrap(err, "unable to update filter dimension"),
+				logData: logData,
+			},
+		)
+		return
+	}
+
+	responseDimension := dimensionItem{}
+	responseDimension.fromDimension(updatedDimension, api.cfg.FilterAPIURL, filterID)
+
+	w.Header().Set("ETag", newETag)
+
+	api.respond.JSON(ctx, w, http.StatusOK, responseDimension)
 }
 
 func (api *API) putFilter(w http.ResponseWriter, r *http.Request) {
--- a/api/filters.go
+++ b/api/filters.go
@@ -423,11 +423,11 @@ func (api *API) updateFilterDimension(w http.ResponseWriter, r *http.Request) {
 		"dimension_name": dimensionName,
 	}
 
-	updatedDimension := model.Dimension{
+	dim := model.Dimension{
 		Options: []string{},
 	}
 
-	if err := api.ParseRequest(r.Body, &updatedDimension); err != nil {
+	if err := api.ParseRequest(r.Body, &dim); err != nil {
 		api.respond.Error(
 			ctx,
 			w,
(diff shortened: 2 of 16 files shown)


api/error.go
```diff
@@ -42,6 +42,24 @@ func (e Error) LogData() map[string]interface{} {
 	return e.logData
 }
 
+// notFoundErr satisfies the `errUnavailable` interface
```
ChrysmOre, line 45: I think it would be better to keep to how the error types are implemented in the mongo package, i.e. add the NotFound tags and functions to the main error type. Not only will it be consistent but this way you can return logData at the same time etc 
```suggestion
// Error is the packages error type
type Error struct {
	err     error
	message string
	logData map[string]interface{}
	notFound bool
	badRequest bool
}

func (e err) NotFound() bool{
  return e.notFound
}
```
etc

api/filters.go
```diff
@@ -420,6 +420,72 @@ func (api *API) addFilterDimension(w http.ResponseWriter, r *http.Request) {
...
+	filterID := chi.URLParam(r, "id")
+	dimensionName := chi.URLParam(r, "name")
+
+	logData := log.Data{
+		"filter_id":      filterID,
+		"dimension_name": dimensionName,
+	}
+
+	updatedDimension := model.Dimension{
+		Options: []string{},
```
ChrysmOre, line 434: Do you need to initialise this slice here?

api/filters.go
```diff
@@ -420,6 +420,72 @@ func (api *API) addFilterDimension(w http.ResponseWriter, r *http.Request) {
...
+			statusCode(err),
+			Error{
+				err:     fmt.Errorf("unable to update filter dimension: %w", err),
+				logData: logData,
+			},
+		)
+		return
+	}
+
+	w.Header().Set("ETag", newETag)
```
ChrysmOre, line 484: probably best to use the const eTagHeader here.. Actually not sure if it's been added yet, might be in an upcoming PR

api/filters.go
```diff
@@ -420,6 +420,72 @@ func (api *API) addFilterDimension(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	newETag, err := api.store.UpdateFilterDimension(ctx, filterID, dimensionName, updatedDimension, eTag)
+	if err != nil {
+		api.respond.Error(
+			ctx,
+			w,
+			statusCode(err),
+			Error{
+				err:     fmt.Errorf("unable to update filter dimension: %w", err),
```
ChrysmOre, line 477: We should be using errors.Wrap rather than fmt.Errorf

api/filters.go
```diff
@@ -557,6 +623,34 @@ func (api *API) isValidDatasetDimensions(w http.ResponseWriter, ctx context.Cont
...
+			errors.Wrap(err, "error retrieving filter"),
+		}
+	}
+
+	foundDimensions, err := api.ctblr.SearchDimensions(ctx, cantabular.SearchDimensionsRequest{
+		Dataset: filter.PopulationType,
+		Text:    dimensionName,
+	})
+
+	if err != nil {
```
ChrysmOre, line 641: better to keep error checks immediately after generated (even if it was a multi-line call)

datastore/mongodb/error.go
```diff
@@ -45,3 +46,9 @@ func (e *er) Conflict() bool {
 func (e *er) Unavailable() bool {
 	return e.unavailable
 }
+
+// BadRequest satisfies the errBadRequest interface and allows other packages
+// to recall metadata about the error thrown
+func (e *er) BadRequest() bool {
```
ChrysmOre, line 52: We don't want the mongo package to know anything specific to http directly, even though the existing stuff translates pretty well it's not a 1-to-1

datastore/mongodb/filters.go
```diff
@@ -141,3 +142,88 @@ func (c *Client) AddFilterDimension(ctx context.Context, fID string, dimension m
...
+		return "", errors.Wrap(err, "error while locking filters collection: %w")
+	}
+
+	defer col.unlock(ctx, lockID)
+
+	filter, err := c.GetFilter(ctx, filterID)
+	if err != nil {
+		return "", &er{
+			err:        errors.Wrapf(err, "unable to fetch filter (%s)", filterID),
+			badRequest: true,
```
ChrysmOre, line 160: I think it would be better to follow the exising GetFilter error logic here
```suggestion
		err := &er{
			 errors.Wrapf(err, "unable to fetch filter (%s)", filterID)
		}
		if errors.Is(err, mongodb.ErrNoDocumentFound) {
			err.notFound = true
		}
		return nil, err
```

datastore/mongodb/filters.go
```diff
@@ -141,3 +142,88 @@ func (c *Client) AddFilterDimension(ctx context.Context, fID string, dimension m
...
+	updateFilter := bson.M{
+		"$set": bson.M{
+			"etag":         filter.ETag,
+			"dimensions.$": dimension,
+		},
+	}
+
+	result, err := c.conn.Collection(col.name).Update(ctx, selectFilter, updateFilter)
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to update filter (%s) dimension (%s)", filterID, dimensionName)
```
ChrysmOre, line 206: Where you're using a lot of wrapf etc, would it be better to move some of that information into logData?

features/steps/steps.go
```diff
@@ -190,6 +224,38 @@ func (c *Component) anETagIsReturned() error {
...
+		return errors.Wrap(err, "failed to retrieve filter")
+	}
+
+	hash, err := response.Hash(nil)
+	if err != nil {
+		return errors.Wrap(err, "unable to hash stored filter")
+	}
+
+	if eTag != hash {
+		return fmt.Errorf("ETag header did not match, expected %s, got %s", hash, eTag)
```
ChrysmOre, line 249: We should be using errors.Wrap, wrapf, errorf etc rather than fmt.Errorf

features/steps/steps.go
```diff
@@ -231,19 +297,78 @@ func (c *Component) theMaximumLimitIsSetTo(val int) error {
...
+	}
+
+	return nil
+}
+
+// cantabularSearchReturnsTheseDimensions sets up a stub response for the `SearchDimensions` method.
+func (c *Component) cantabularSearchReturnsTheseDimensions(datasetID, dimension string, docs *godog.DocString) error {
+	var response cantabular.GetDimensionsResponse
+	if err := json.Unmarshal([]byte(docs.Content), &response); err != nil {
+		return err
```
ChrysmOre, line 337: wrap error?

ChrysmOre (changes requested):

datastore/mongodb/filters.go
jayfreestone: When we refined the story we agreed to stick to Fran's Swagger spec and return a `400` if the filter wasn't found. We can change it (definitely wasn't set in stone), but if we _do_ want to return a `400` then we need to differentiate between a bad request/not found in some way on the Mongo error.

Could be a bool like `failedValidation` rather than `badRequest` in order to make it less of a direct mapping, although we would still need to add an additional error interface/check. 🤔

Let me know what you prefer. I agree that it's not great having response terminology leak into the DB layer, it was just driven by the spec 👍

api/filters.go
jayfreestone: Doesn't appear to be in yet. Can do a search and replace once the change goes in — think there are a couple of places just using a string.

api/filters.go
jayfreestone: This was to avoid absent options getting stored/serialised as `null`. Happy to remove if you prefer.

api/error.go
jayfreestone: Makes sense to be consistent, have updated 👍. My qualm with a shared type and multiple flags is that it's usually the result of trying to represent a union — it's unlikely you'd want an error that is both `notFound` and `badRequest`, but unfortunately the type system permits it if you go this route.

ChrysmOre: Yeah it does leave that room for programmer error.. At the moment the way it resolves if does happen that there's a hierarchy between the codes. It could be that we could implement an enum for the underlying values rather than bools, so long as we're still using the individual interfaces to determine the result.

datastore/mongodb/filters.go
jayfreestone: `badRequest` addition to the Mongo package removed, since the validation does indeed handle the `400` response code 👍

jayfreestone:
@ChrysmOre Fixed the conflicts with the filter model changes, ready for a re-review when you get a sec.

api/filters.go
```diff
@@ -414,6 +414,75 @@ func (api *API) addFilterDimension(w http.ResponseWriter, r *http.Request) {
...
+
+	// The new dimension won't be present on the dataset (i.e. only `City` will be present, not `Country`),
+	// so instead of validating it against the existing Version, we check to see if the dimension exists in Cantabular.
+	if err := api.validateCantabularDimensionExists(ctx, filterID, updatedDimension.Name); err != nil {
+		api.respond.Error(
+			ctx,
+			w,
+			statusCode(err),
+			Error{
+				err:     errors.Wrapf(err, "error searching for dimension %s in cantabular", updatedDimension.Name),
```
ChrysmOre, line 452: We don't want to be exposing the word Cantabular (or any other details of our tech stack) to the public. Also we should be trying to keep error messages static, and add the dynamic data to the log data. Here we already have the dimension name in the so it may not be needed at all.

api/filters.go
```diff
@@ -414,6 +414,75 @@ func (api *API) addFilterDimension(w http.ResponseWriter, r *http.Request) {
...
+			statusCode(err),
+			Error{
+				err:     errors.Wrap(err, "unable to update filter dimension"),
+				logData: logData,
+			},
+		)
+		return
+	}
+
+	responseDimension := dimensionItem{}
```
ChrysmOre, line 478: Couple of things here.. Can we keep to the `updateFilterDimensionResponse` naming convention here too? And also a smaller thing but does add up over time, it's really in the spirit of Go to keep to shorter variable names as much as possible. Here `resp` would be plenty (and in line with what we call it elsewhere) and earlier you might want to consider if it's necessary to specify `updatedDimension` when maybe `dimension` or `dim` would do.

api/filters.go
```diff
@@ -414,6 +414,75 @@ func (api *API) addFilterDimension(w http.ResponseWriter, r *http.Request) {
...
+	responseDimension.fromDimension(updatedDimension, api.cfg.FilterAPIURL, filterID)
+
+	w.Header().Set("ETag", newETag)
```
ChrysmOre, line 481: I think the ETag const has been added now?

api/filters.go
```diff
@@ -602,6 +671,36 @@ func (api *API) isValidDatasetDimensions(w http.ResponseWriter, ctx context.Cont
...
+		Dataset: filter.PopulationType,
+		Text:    dimensionName,
+	})
+	if err != nil {
+		return errors.Wrap(err, "error in cantabular response")
+	}
+
+	if len(foundDimensions.Dataset.Variables.Search.Edges) == 0 {
+		return Error{
+			err:      fmt.Errorf("no dimensions in response (%+v)", foundDimensions),
```
ChrysmOre, line 696: Should be errors.Wrap (also moving dimension list to log data)

features/steps/steps.go
```diff
@@ -194,19 +263,78 @@ func (c *Component) theMaximumLimitIsSetTo(val int) error {
...
-	ctx := context.Background()
 	var filters []model.Filter
 
 	err := json.Unmarshal([]byte(docs.Content), &filters)
 	if err != nil {
 		return errors.Wrap(err, "failed to unmarshall")
 	}
 
+	if err := c.insertFilters(filters); err != nil {
+		return err
```
ChrysmOre, line 274: Probably good to wrap this error here

ChrysmOre (approved):

PR #47, opened 2022-04-05
Feature/dataset json

s-xton:
### What

Added get dataset JSON endpoint implementation 
### How to review

Review the code, run tests and test against a known dataset

### Who can review

anyone


api/dataset.go
```diff
@@ -2,38 +2,218 @@ package api
...
+		api.getDefaultDatasetJSON(ctx, w, r)
 	} else {
-		response, err = api.getGeographyDatsetJSON(ctx, r)
+		api.getGeographyDatsetJSON(ctx, r)
+	}
+}
+
+func (api *API) getDefaultDatasetJSON(ctx context.Context, w http.ResponseWriter, r *http.Request) {
+	params, err := api.getDatasetParams(ctx, r)
+
```
ChrysmOre, line 37: Please check errors immediately after they're created rather than leaving blank lines between them

api/dataset.go
```diff
@@ -2,38 +2,218 @@ package api
...
+	if err != nil {
+		api.respond.Error(ctx, w, http.StatusNotFound, err)
```
ChrysmOre, line 39: StatusNotFound? It seems this function can return a lot of errors that would be inappropriate to return a 404 for.
Also we have a convention of wrapping our errors with context at each level so the error messages are meaningful.
For example here we would do:
```
api.Respond.Error(ctx, w, statusCode(err), errors.Wrap(err, "failed to get dataset params" )
``` 
And then do the equivalent for all errors at all levels of the call stack. This avoids completely mysterious error messages like `invalid numeral at line 6` which might end up something like `failed to get dataset params: invalid dataset id: invalid numeral at line 6`

api/dataset.go
```diff
@@ -2,38 +2,218 @@ package api
...
+		return
+	}
+
+	geoDimensions, err := api.getGeographyTypes(ctx, params.basedOn)
+	if err != nil {
+		api.respond.Error(ctx, w, http.StatusInternalServerError, err)
```
ChrysmOre, line 45: Please don't just return status.InternalServerError, we have set up a system that can reflexively determine statusCodes from information embedded in the errors. As a default you should return statusCode(err), which will attempt to determine the statusCode and default to status 500. Here the error returned by cantabular will return the actual status code. Other status codes are achieved by returning errors that conform to interfaces found in package `errors`

api/dataset.go
```diff
@@ -2,38 +2,218 @@ package api
...
 
-func (api *API) getDefaultDatasetJSON(ctx context.Context, r *http.Request) (*getDatasetJsonObservationsResponse, error) {
+func (api *API) getGeographyDatsetJSON(ctx context.Context, r *http.Request) (*getDatasetJsonObservationsResponse, error) {
 	return &getDatasetJsonObservationsResponse{}, nil
 }
 
-func (api *API) getGeographyDatsetJSON(ctx context.Context, r *http.Request) (*getDatasetJsonObservationsResponse, error) {
-	return &getDatasetJsonObservationsResponse{}, nil
+func (api *API) toGetDatasetJsonResponse(dimensionMap optionsMap, query *cantabular.StaticDatasetQuery) (*getDatasetJsonObservationsResponse, error) {
+	var dimensions []getDatasetJsonResponseDimension
```
ChrysmOre, line 78: Sorry I missed the initial PR but there seems to be a lot of extra structs with obtuse names that should be organised a lot better. A lot of structs are duplicates of what we already have. We have quite a particular way of keeping things organised and I would like to keep it that way, otherwise things will quickly become unbearable to work with. I will give some concrete examples below

api/dataset.go
```diff
@@ -2,38 +2,218 @@ package api
...
+	}
+
+	dimensions, err := api.datasets.GetVersionDimensions(ctx, "", "", "", params.id, params.edition, params.version)
+
+	if err != nil {
+		return nil, err
+	}
+
+	if dimensions.Items.Len() == 0 {
+		return nil, errors.New("invalid dimensions length")
```
ChrysmOre, line 154: Would specifying length equalling zero be more informative here?

api/dataset.go
```diff
@@ -2,38 +2,218 @@ package api
...
+	for _, dimension := range res.Dataset.RuleBase.IsSourceOf.Edges {
+		geoDimensions = append(geoDimensions, dimension.Node.Name)
+	}
+
+	return geoDimensions, nil
+}
+
+func (api *API) sortGeography(geoDimensions []string, dimensions []string) ([]string, bool) {
+	foundGeography := false
+	sortedDimensions := []string{}
```
ChrysmOre, line 194: It's bad practice to initialise slices before they're used unless necessary to avoid pre-emptive memory allocation. You should initialise the slices using `var` where possible.

api/dataset.go
ChrysmOre: In contract.go:
- `getDatasetJsonResponseLink`, and `getDatasetJsonResponseDimensionOption` are unnecessary, please use the `model.Link` struct as we have elsewhere.
- I'm wondering why `getDatasetJsonResponseVersionLink` has a `version` field instead of an `id`? Everywhere else through the API versions are referred to purely by an HREF as some links are. If this is definitely correct then it would be better to keep the naming simpler i.e.`VersionLink`. We don't need each struct to only belong to one other struct, we want to re-use things as much as possible.
- `getDatasetJsonObservationsResponse` is not in line with our naming convention. If the endpoint/handler is called `getDatasetJSON` then the response should be called `getDatasetJSONResponse`. It should be a 1:1 match.
- For `getDatasetJsonResponseLinks` and `getDatasetJsonResponseDimension` you don't need to specify 'get' 
 and 'response'. `DatasetJSONLinks` and `DatasetJSONDimension` would be enough. Although `DatasetJSONDimensionItem` might be more in line with other naming in the package. Either way, we want to strive to keep type and variable names as short/simple as possible while retaining the useful information.

ChrysmOre (changes requested):
A few comments plus some general:
There doesn't seem to be any use of logData anywhere, there is probably a lot of places this would help. Likewise we usually don't want to return the intimate details of every error to the user. Please use the errorMessage feature of the api package error type to tailor error messages that are suitable to be returned to the public (Presuming this endpoint will be public facing?).
I can't see the route being added to the api anywhere, but I have a couple of questions: I see in a comment in contract.go it suggests you're aiming to have this at `/flex/datasets/...`. Unless something has changed without me knowing the use of the `/flex` prefix was ruled out some time ago. I'm not sure how you plan to set the routing up but the `/datasets` prefix is obviously taken by dp-dataset-api. Are you plannig to call this endpoint from there?

api/dataset.go
s-xton: Updated endpoint path

api/dataset.go
```diff
@@ -3,37 +3,276 @@ package api
...
+	params.datasetLink = datasetItem.Links.Self
+	params.basedOn = datasetItem.IsBasedOn.ID
+
+	if datasetItem.DatasetDetails.Type != "cantabular_flexible_table" {
+		return nil, errors.New("invalid dataset type")
+	}
+
+	versionItem, err := api.datasets.GetVersion(ctx, "", "", "", "", params.id, params.edition, params.version)
+	if err != nil {
+		return nil, err
```
ChrysmOre, line 197: Can we make sure we're wrapping errors each time we're returning? Just a brief description of what's failed will do, it leads to much more useful error messages e.g.
```
return nil, errors.Wrap(err, "failed to get version")
```

ChrysmOre (approved):

PR #48, opened 2022-04-06
Feature/get filter outputs method added

saminahbab:
### What

- Added GetFilterOutput method to service 
- Added interface methods, testing stubs for filterOutput retrieval and a query to recieve
- Added component tests, testing for 500, 200 and 404 codes 
- Added clearer step for failing mongo database.

### How to review

Run `make test` and `make test-component` 

### Who can review

anyone from Team B can review


features/put_filter_outputs.feature
```diff
@@ -11,9 +11,9 @@ Feature: Put Filter Outputs Private Endpoints Enabled
         "filter_id": "94310d8d-72d6-492a-bc30-27584627fil1",
         "instance_id": "94310d8d-72d6-492a-bc30-27584627inst1",
         "state": "published",
-        "downloads": 
+        "downloads":
```
saminahbab, line 14: These formatting commits, I believe, are from an earlier merge, and are being brought back in through this PR. I thought it was worthwhile as they are removing redundant spaces and tabs.

api/filter_outputs.go
```diff
@@ -1,14 +1,63 @@
...
 
+func (api *API) getFilterOutput(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	fID := chi.URLParam(r, "filter-output-id")
+
+	var filterOutput *model.FilterOutput
+
+	filterOutput, err := api.store.GetFilterOutput(ctx, fID)
+	if err != nil {
+
```
ChrysmOre, line 22: No need to leave blank lines at the beginning of code blocks. In the long run it's more legible to keep consistent.

api/filter_outputs.go
```diff
@@ -1,14 +1,63 @@
...
+		status := http.StatusNotFound
+		message := "filter output not found"
+
```
ChrysmOre, line 25: Ok this stuff is a little unnecessary, if you look at some of the existing mongo functions you'll see we add tags like `notFound` etc to the errors there, then here we use the statusCode(err) function to determine which statusCode to return. The default will be 500. So here you can just do:
```
if err != nil{
  		api.respond.Error(
			ctx,
			w,
			statusCode(err),
			Error{
				err: errors.Wrap(err, "failed to get filter output),
				logData: log.Data{
					"id": fID,
				},
			},
		)
}
```

ChrysmOre: Also in general it's considered bad practice to inspect the error strings to determine what to do with errors. It's better to use interfaces to determine behaviour (as we have set up, you'll see more further down)

datastore/mongodb/filter_outputs.go
```diff
@@ -9,6 +9,19 @@ import (
...
 )
 
+// GetFilterOutput
+func (c *Client) GetFilterOutput(ctx context.Context, filterID string) (*model.FilterOutput, error) {
+	var filterOutput model.FilterOutput
+
+	coll := c.collections.filterOutputs
+
+	if err := c.conn.Collection(coll.name).FindOne(ctx, bson.M{"id": filterID}, &filterOutput); err != nil {
+		return nil, err
```
ChrysmOre, line 19: So here with the errors if you check the example in getFilter:
```
if err = c.conn.Collection(col.name).FindOne(ctx, bson.M{"filter_id": fID}, &f); err != nil {
		err := &er{
			err: errors.Wrap(err, "failed to find filter"),
		}
		if errors.Is(err, mongodb.ErrNoDocumentFound) {
			err.notFound = true
		}
		return nil, err
	}
```
you can see how we're annotating the error with information here which we then use in the handler to determine what to do with it

datastore/mongodb/filter_outputs.go
```diff
@@ -29,6 +42,7 @@ func (c *Client) CreateFilterOutput(ctx context.Context, f *model.FilterOutput)
 
 // UpdateFilterOutput creates/updates a FilterOutputs in the CantabularFilters collection
 func (c *Client) UpdateFilterOutput(ctx context.Context, f *model.FilterOutput) error {
+
```
ChrysmOre, line 45: Again, tempting as  it may be try to avoid adding blank lines at the beginning of functions.. Once you stick to one method for long enough it will start to always look right

ChrysmOre (changes requested):

api/filter_outputs.go
saminahbab: resolved, and noted

datastore/mongodb/filter_outputs.go
saminahbab: removed

api/filter_outputs.go
saminahbab: @ChrysmOre  understood and noted. The recent commit checks with `statusCode(err)` as you ask. However, the error codes are now 500 for both of the failed cases that I have tested, which I assume is okay. I ammended the tests to reflect this. 

Also I believe the use of `statusCode(err)` isn't consistent in some of the methods, where instead `http.Status<codetype>` is used instead. This probably can be cleaned up in the future.

saminahbab: @ChrysmOre this has been resolved. `statusCode(err)` is now being used, with 404s and 500s returned.

datastore/mongodb/filter_outputs.go
saminahbab: @ChrysmOre  not sure why this is not being marked as outdated, but if you look at https://github.com/ONSdigital/dp-cantabular-filter-flex-api/blob/feature/get-filter-outputs/datastore/mongodb/filter_outputs.go you will find that the changes you have requested have been implemented.

ChrysmOre: Outdate just means the file/code has changed since I made the comment

ChrysmOre (commented):
LGTM, but I'm going to ask you refrain from merging just yet until the data model updates have been merged.. You will need to go and do some merging once they are because some of the fundamental models you use here will be affected

api/contract.go
```diff
@@ -95,6 +95,17 @@ func (r *updateFilterOutputRequest) Valid() error {
 	return nil
 }
 
+// getFilterOutputResponse is the response body for GET/filter-outputs/{id}
+type getFilterOutputResponse struct {
+	FilterID         string                  `json:"filter_id"`
+	InstanceID       string                  `json:"instance_id"`
+	DimensionListURL string                  `json:"dimension_list_url"`
```
ChrysmOre, line 102: Pretty sure dimensionListURL is no longer part of the spec.. I see the new swagger spec hasn't been approved and merged yet.. probably a good idea to do that

features/put_filter_outputs.feature
saminahbab: Resolving as assuming this comment has been seen now.

api/filter_outputs.go
```diff
@@ -9,6 +10,42 @@ import (
...
+				logData: log.Data{
+					"id": fID,
+				},
+			},
+		)
+		return
+	}
+
+	resp := getFilterOutputResponse{
+		filterOutput.FilterID,
```
ChrysmOre, line 36: If I'm not mistaken, the getFilterOutput response equates to simply a FilterOutput resource. The simplest way to do this is to define the GetFilterOutputResponse as simply haiving the FilterOutput embedded into it, and you can assign it as `resp := getFilterOutputResponse{FilterOutput: filterOutput}`. This is done a few times in the other endpoints for an example

api/filter_outputs.go
```diff
@@ -9,6 +10,42 @@ import (
...
+		filterOutput.InstanceID,
+		fmt.Sprintf("%s/filter-outputs/%s", api.cfg.BindAddr, filterOutput.FilterID),
+		filterOutput.Events,
+		filterOutput.Links,
+		filterOutput.Downloads,
+		filterOutput.State,
+	}
+
+	api.respond.JSON(ctx, w, http.StatusOK, resp)
+
```
ChrysmOre, line 46: Extra blank line

api/contract.go
saminahbab: New swagger spec has been merged

api/filter_outputs.go
saminahbab: In light of the new swagger sped, this now makes sense, I have done so in the last commit, and amended feature tests to reflect the new response. I will mark as resolved.

api/filter_outputs.go
saminahbab: removed. I will resolve

features/get_filter_output.feature
```diff
@@ -0,0 +1,200 @@
...
+    """
+    And the HTTP status code should be "200"
+
+  Scenario: Filter Output not found
+    When I GET "/filter-outputs/not-exist"
+
+    Then I should receive the following JSON response:
+    """
+    {
+      "errors": ["failed to get filter output: mongo: no documents in result"]
```
ChrysmOre, line 174: I don't think we should be exposing the details of our database tech to the public.. Elsewhere for a 404 we just return 'filter not found' with the 404 status code. You can do this with the errorMessage field of the error in the handler or elsewhere

ChrysmOre (dismissed):

ChrysmOre (approved):

api/api.go
```diff
@@ -59,6 +59,7 @@ func (api *API) enablePublicEndpoints() {
 	api.Router.Get("/filters/{id}/dimensions", api.getFilterDimensions)
 	api.Router.Post("/filters/{id}/dimensions", api.addFilterDimension)
 	api.Router.Get("/flex/datasets/{dataset_id}/editions/{edition}/versions/{version}/json", api.getDatasetJSON)
+	api.Router.Get("/filter-outputs/{filter-output-id}", api.getFilterOutput)
```
j-s-rawat, line 62: might be good to change it to 'filter_output_id' to keep it consistent

api/api.go
```diff
@@ -77,7 +78,10 @@ func (api *API) enablePrivateEndpoints() {
 	r.Post("/filters/{id}/submit", api.submitFilter)
 	r.Get("/filters/{id}/dimensions", api.getFilterDimensions)
 	r.Post("/filters/{id}/dimensions", api.addFilterDimension)
+
+	r.Get("/filter-outputs/{filter-output-id}", api.getFilterOutput)
```
j-s-rawat, line 82: same as above

api/filter_outputs.go
```diff
@@ -9,6 +9,36 @@ import (
...
+				message: "failed to get filter output",
+				logData: log.Data{
+					"id": fID,
+				},
+			},
+		)
+		return
+	}
+
+	resp := getFilterOutputResponse{
```
j-s-rawat, line 35: Can we directly use *filterOutput in the api.respond.JSON without the need of this temp variable

PR #53, opened 2022-04-08
update submitFilterResponse and exportStart event

ChrysmOre:
### What

- Updated Submit Filter endpoint to return and forward FilterOutputID instead of FilterID
- Couple of small tidy-up changes

### How to review

Check changes make sense, tests pass

### Who can review

Anyone


PauloPortugal (dismissed):
So we completely miss the context of the previous work?
This seems like a tidying up work.
It would be good to understand why we got here.

PauloPortugal (approved):

PauloPortugal (approved):

PR #55, opened 2022-04-11
Fix/wrong kafka test

saminahbab:
### What

Fixed a test that was testing for the wrong condition. 

The Kafka consumer in one of the tests was testing for the absence of an event in the topic, rather than the presence of an event. This was changed around.

### How to review

run `make test-component`.

### Who can review

anyone from team b can review


PauloPortugal (approved):
LGTM

PR #60, opened 2022-04-21
Add ID/Label to stored filters (#5664)

jayfreestone:
### What

2022-04-21 at 11 07 06

Now stores two additional fields on a filter dimensions: ID and label.

Previously we assumed that the `name` field on a filter was sufficient, since it would contain the selection for a dimension (e.g. `name: "city"`). Our new understanding is that the underlying name should remain unchanged (e.g. consistently `name: "geography"`), which means that we also need to store the currently selected choice (`id`). The label is included as a convenience for the frontend and any other consumers.

This is one step towards [5664](https://trello.com/c/3TbBPTvY), and will enable the frontend to query for/update a specific filter dimension using a consistent machine-readable name (once the import process has also been updated).

### How to review

Ensure the change makes sense, both conceptually/schema wise and in implementation.

### Who can review

Anyone, but probably Team B.

### DIFF
--- a/api/filters.go
+++ b/api/filters.go
@@ -371,7 +371,9 @@ func (api *API) addFilterDimension(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if err := api.store.AddFilterDimension(ctx, fID, req.Dimension); err != nil {
+	dim := hydrateDimensions([]model.Dimension{req.Dimension}, v.Dimensions)[0]
+
+	if err := api.store.AddFilterDimension(ctx, fID, dim); err != nil {
 		api.respond.Error(
 			ctx,
 			w,
@@ -386,7 +388,7 @@ func (api *API) addFilterDimension(w http.ResponseWriter, r *http.Request) {
 
 	var resp addFilterDimensionResponse
 	resp.dimensionItem.fromDimension(
-		req.Dimension,
+		dim,
 		api.cfg.FilterAPIURL,
 		fID,
 	)
--- a/api/filters.go
+++ b/api/filters.go
@@ -83,7 +83,7 @@ func (api *API) createFilter(w http.ResponseWriter, r *http.Request) {
 				ID: strconv.Itoa(v.Version),
 			},
 		},
-		Dimensions:        req.Dimensions,
+		Dimensions:        hydrateDimensions(req.Dimensions, v.Dimensions),
 		UniqueTimestamp:   api.generate.UniqueTimestamp(),
 		LastUpdated:       api.generate.Timestamp(),
 		Dataset:           *req.Dataset,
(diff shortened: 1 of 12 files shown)


api/filters.go
```diff
@@ -429,11 +432,14 @@ func (api *API) updateFilterDimension(w http.ResponseWriter, r *http.Request) {
 		"dimension_name": dimensionName,
 	}
 
-	dim := model.Dimension{
-		Options: []string{},
+	updateReq := updateFilterDimensionRequest{
```
ChrysmOre, line 435: Unless there is another type of request to differentiate from just `req` is fine/preferable

api/filters.go
```diff
@@ -676,12 +686,12 @@ func (api *API) isValidDatasetDimensions(w http.ResponseWriter, ctx context.Cont
 	return true
 }
 
-// validateCantabularDimensionExists checks that dimension exists in Cantabular by searching for it.
+// getCantabularDimension checks that dimension exists in Cantabular by searching for it.
 // If the dimension doesn't exist, or couldn't be retrieved, an error is returned.
-func (api *API) validateCantabularDimensionExists(ctx context.Context, filterID, dimensionName string) error {
+func (api *API) getCantabularDimension(ctx context.Context, filterID, dimensionName string) (gql.Node, error) {
```
ChrysmOre, line 691: Any reason not return a *gql.Node (and then return nil on error)?

features/steps/steps.go
```diff
@@ -265,9 +267,32 @@ func (c *Component) privateEndpointsAreNotEnabled() error {
 	return nil
 }
 
-func (c *Component) theDocumentInTheDatabaseShouldBe(id string, doc *godog.DocString) error {
-	// TODO: implement step for verifying documents stored in Mongo. No prior
-	// art of this being done properly in ONS yet so save to be done in future ticket
+// theDocumentInTheDatabaseShouldMatch checks if the filter in the DB matches the one provided.
+// Timestamps and the ETag are ignored.
+func (c *Component) theDocumentInTheDatabaseShouldMatch(fID string, doc *godog.DocString) error {
```
ChrysmOre, line 272: Whey, finally someone got round to it! :)

features/steps/steps.go
```diff
@@ -265,9 +267,32 @@ func (c *Component) privateEndpointsAreNotEnabled() error {
...
+
+	var response model.Filter
+	if err := c.store.Conn().Collection(col).FindOne(ctx, bson.M{"filter_id": fID}, &response); err != nil {
+		return errors.Wrap(err, "failed to retrieve filter")
+	}
+
+	// ETags are compared by a separate assertion
+	response.ETag = ""
+	// Don't compare time, since it will be non-deterministic
+	response.LastUpdated = time.Time{}
```
ChrysmOre, line 289: These fields should be predetermined in the tests as we're using the mock generator? (see features/mock/generator.go)

api/filters.go
jayfreestone: Don't have a strong pref so have changed 👍

I guess generally if a function is responsible for actually creating a type (e.g. making a request and deserialising the response), then I tend to return the type rather than a pointer. I can see an argument for either, but I'm now so used to Rust's stricter ownership rules 🦀.

features/steps/steps.go
jayfreestone: Updated to use your revised version of the step method.

ChrysmOre (approved):

ChrysmOre (approved):

PR #63, opened 2022-04-22
Stop passing a reference to event.ExportStart

PauloPortugal:
### What

There is no need to pass a pointer/reference.
A copy will suffice and is more efficient for such a small struct.

### How to review

This PR

### Who can review

Anyone.


andre-urbani (approved):
LGTM

PR #70, opened 2022-05-13
Upgrade/clients upgrade

saminahbab:
### What

Upgrading the api-clients-library to v1.132
Updated some of the interfaces to be in line with the client library.

### How to review

run the relevant testing steps 

### Who can review

Team b


mr-nick17 (dismissed):
LGTM

PR #71, opened 2022-05-18
allow public or private to be empty when updating filter ouput downloads

ChrysmOre:
### What

Updated validation for updating filter outputs to allow 'public' and 'private' fields to be updated individually.

### How to review

Check change makes sense

### Who can review

Anyone


PauloPortugal (dismissed):

PauloPortugal (approved):
LGTM

PR #80, opened 2022-06-23
add POST /filter/{id}/dimensions/{dimension}/options/{option}

ChrysmOre:
### What

Added POST /filters/{id}/dimensions/{dimension}/options/{option}

Adds a specific option to a dimension on an already existing filter.

Straightforward handler, performs a few checks (does dimension exist, does option already exist, is it a valid value) before adding the option.

### How to review

Check changes makes sense, tests pass

### Who can review

Anyone


features/steps/steps.go
```diff
@@ -107,7 +107,7 @@ func (c *Component) filterOutputIsInDatastore(expectedOutput *godog.DocString) e
 
 	err := json.Unmarshal([]byte(expectedOutput.Content), &expected)
 	if err != nil {
-		return errors.Wrap(err, "failed to unmarshall provided filterOutput")
+		return fmt.Errorf("failed to unmarshall provided filterOutput: %w", err)
```
PatrickConnollyONS, line 110: I had a question about the use of fmt.Errorf() in this file: I'm assuming you didn't want githum.com/pkg/errors/Wrap() to avoid the storing of the stack frame. But would it not be better to use githum.com/pkg/errors/WrapMessage() instead of fmt.Errorf() - it's confusing switching between the use of different packages for error creation, especially since githum.com/pkg/errors has an Errorf() function which does not wrap an error, and so compounds teh confusion in the switch to fmt.Errorf()

PatrickConnollyONS (dismissed):
Hi Joshua, the code looks good and all test pass - I just had a general question/comment as to the use of the fmt.Errorf() in steps.go.

api/filter_dimensions.go
```diff
@@ -222,6 +223,7 @@ func (api *API) updateFilterDimension(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// ID/name is provided by the request, but the label is provided by Cantabular.
+	// TODO: WHere is this used/necessary?
```
PauloPortugal, line 226: Should we just remove it, if this is not used?

ChrysmOre: I can't work out how, but the tests to break if I remove it

api/filters.go
```diff
@@ -298,6 +297,7 @@ func (api *API) getFilter(w http.ResponseWriter, r *http.Request) {
 	api.respond.JSON(ctx, w, http.StatusOK, resp)
 }
 
+// TODO: what is this unfinished function?
```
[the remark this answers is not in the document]
ChrysmOre: Going to leave it for further investigation after this morning's discussion

features/steps/steps.go
ChrysmOre: Ah I just remembered why I changed it actually, it's not just to remove the stack trace it was that the output wasn't being shown properly. I think it's something to do with the way the GoDog packages work but only the wrapping message was being shown, not the underlying error.

PauloPortugal (approved):

PR #81, opened 2022-06-24
Feature/5729 dimension options

saminahbab:
[Trello Ticket Here](https://trello.com/c/QhqBbglU/5729-new-endpoint-in-cantabular-filter-flexi-api-get-filters-filterid-dimensions-dimension-options)
### What

Add Route `filters/<ID>/dimensions/<NAME>/options` to api. 
Add component tests to test various test cases. 
Add mongo method to return paginated options given filter and dimension name. 
Fill necessary interface methods for database and handler 

### How to review

Run `make test` to see no regressions, and `make test-component` for no regressions and added component tests

### Who can review

team b


api/filter_dimensions.go
```diff
@@ -0,0 +1,372 @@
...
+	}
+
+	var resp addFilterDimensionResponse
+	resp.dimensionItem.fromDimension(
+		dim,
+		api.cfg.FilterAPIURL,
+		fID,
+	)
+
+	w.Header().Set(eTagHeader, b)
```
saminahbab, line 174: Am not totally sure I understand this part of the journey. I may need some guidance on whether the retrieval and the setting of the eTag is correct.

api/options.go
```diff
@@ -0,0 +1,110 @@
...
+		Items: parseFilterDimensionOptions(options, filterID, dimensionName),
+		paginationResponse: paginationResponse{
+			Limit:      pageLimit,
+			Offset:     offset,
+			Count:      len(options),
+			TotalCount: totalCount,
+		},
+	}
+
+	w.Header().Set(eTagHeader, eTag)
```
saminahbab, line 81: I am not sure that I understand fully the process of retrieving and setting this eTag. I would like some guidance on this.

api/filter_dimensions.go
```diff
@@ -0,0 +1,372 @@
+package api
```
saminahbab, line 1: I am not sure how this file has been added to my PR. I have merged in develop twice. It may be from that, or it could be deleted work that now is being folded back in. Please advise.

api/contract.go
```diff
@@ -109,6 +109,19 @@ func (r *putFilterOutputRequest) Valid() error {
 	return nil
 }
 
+type AddOptionResponse struct {
```
ChrysmOre, line 112: Why is this AddOptionResponse? Should it be GetFilterDimensionOptionsItem or FilterDimensionOption etc

api/contract.go
```diff
@@ -109,6 +109,19 @@ func (r *putFilterOutputRequest) Valid() error {
...
+	Option    string     `json:"option"`
+	Self      model.Link `json:"self"`
+	Filter    model.Link `json:"filter"`
+	Dimension model.Link `json:"Dimension"`
+}
+
+// getDimensionOptionsResponse is the response body for GET /filters/{id}/dimensions/{name}/options
+type GetDimensionOptionsResponse struct {
```
ChrysmOre, line 120: Could we name this GetFilterDimensionOptionsResponse so it's distinct from e.g. GetFilterOutputDimensionOptionsResponse etc

api/options.go
```diff
@@ -0,0 +1,110 @@
...
+			errors.Wrap(err, "Bad Request: "),
+		)
+
+	}
+
+	if pageLimit == 0 {
+		// define a reasonable default
+		// in light of bad input
+		// also slice not work with 0
+		pageLimit = 10
```
ChrysmOre, line 34: Generally our default for limit is 20

api/options.go
```diff
@@ -0,0 +1,110 @@
...
+	filterID := chi.URLParam(r, "id")
+	dimensionName := chi.URLParam(r, "name")
+
+	pageLimit, offset, err := getPaginationParams(r.URL, api.cfg.DefaultMaximumLimit)
+	if err != nil {
+		api.respond.Error(
+			ctx,
+			w,
+			http.StatusBadRequest,
+			errors.Wrap(err, "Bad Request: "),
```
ChrysmOre, line 25: You don't need to add `: ` when using errors.Wrap(it adds it for you)

api/options.go
```diff
@@ -0,0 +1,110 @@
...
+		dimensionName,
+		pageLimit,
+		offset,
+	)
+	if err != nil {
+		status := statusCode(err)
+		finalErr := errors.New("internal server error")
+
+		if dperrors.NotFound(err) {
+			status = http.StatusNotFound
```
ChrysmOre, line 54: I'm not sure what the point in this is.. These will assert to these status codes automatically when using statusCode(err) in the call to respond.Error()

api/options.go
```diff
@@ -0,0 +1,110 @@
...
+		if dperrors.BadRequest(err) {
+			status = http.StatusBadRequest
+			finalErr = err
+		}
+
+		api.respond.Error(
+			ctx,
+			w,
+			status,
+			errors.Wrap(finalErr, "Failed to get options"),
```
ChrysmOre, line 66: We shouldn't capitalise error messages

api/options.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func parseFilterDimensionOptions(options []string, filterID, dimensionName string) []AddOptionResponse {
+	responses := make([]AddOptionResponse, 0)
+
+	for _, option := range options {
+		addOptionResponse := AddOptionResponse{
+			Option: option,
+			Self: model.Link{
+				HREF: "/filters/" + filterID + "/dimensions/" + dimensionName + "/options",
```
ChrysmOre, line 93: HREF needs the host `http://etc`, i think we've been using cfg.BindAddr but we should probably actually use filterAPIURL

api/options.go
```diff
@@ -0,0 +1,110 @@
...
+			Limit:      pageLimit,
+			Offset:     offset,
+			Count:      len(options),
+			TotalCount: totalCount,
+		},
+	}
+
+	w.Header().Set(eTagHeader, eTag)
+	api.respond.JSON(ctx, w, http.StatusOK, resp)
+
```
ChrysmOre, line 83: Blank line

datastore/mongodb/error.go
```diff
@@ -52,3 +53,9 @@ func (e *er) Unavailable() bool {
 func (e *er) Forbidden() bool {
 	return e.forbidden
 }
+
+// BadRequest satisfies the badRequest interface and allows other packages
```
ChrysmOre, line 57: Mongo shouldn't know about http requests. If we can't find the dimension with the name we should return a notFound which should then translate to a 404

datastore/mongodb/options.go
```diff
@@ -0,0 +1,65 @@
...
+
+	if options.TotalCount == -1 {
+		return nil, 0, &er{
+			err:        errors.Errorf("failed to find dimension with name (%s)", dimensionName),
+			badRequest: true,
+		}
+	}
+
+	return result[0].Options, result[0].TotalCount, nil
+
```
ChrysmOre, line 64: Blank line

features/get_dimension_options.feature
```diff
@@ -0,0 +1,252 @@
+Feature: Get Filter Dimension Options Private Endpoints
+  Background:
+    Given private endpoints are not enabled
+    And I have these filters:
+    """
+    [
+          {
```
ChrysmOre, line 7: Brace is out of line

features/put_filter.feature
```diff
@@ -120,5 +120,3 @@ Feature: Put Filter Private Endpoints Not Enabled
     }
     """
     And the HTTP status code should be "200"
-
```
ChrysmOre, line 123: Needs a CR (carriage return)

api/contract.go
saminahbab: The swagger spec was defined as such `            $ref: '#/definitions/AddOptionResponse'` ... I was pretty confused by it, but left it as it was defined

api/options.go
saminahbab: If the filter is not found, then the spec says return a 404. If the dimension name is not found within a filter, then says to return a 400 bad request. So in the query that I wrote, there are two ways in which it could fail. The other way it could fail is connection errors. This was the most terse but readable way I could think of to handle the different types of errors coming from the function call

datastore/mongodb/error.go
saminahbab: the rest of the error types in that file are essentially in HTTP request language, so i followed the convention there. 
The swagger spec also defines a missing dimension name as a 400, and a 404 only when filter not found

features/put_filter.feature
saminahbab: I don't see the point of adding one

api/options.go
ChrysmOre: The point I'm getting at is, under the hood respond.Error() does exactly the same thing under the hood, so this is redundant. Passing an error that will assert to dperrors.NotFound == true will automatically set the status code to 404.

api/contract.go
ChrysmOre: Please rename it, it's more important it's not confusing/is consistent in the code than keeping 100% to the name in the swagger spec. Update the name in the swagger spec to match the correct name if anything.

api/options.go
ChrysmOre: The whole point of the eTags is that the caller will send an eTag in the request.. You then need to check that it matches with the stored eTag record. If it doesn't you need to return http.StatusConflict. When updating a record you need to generate a new eTag, store it and return it in the header. What you've done here is just return the eTag the caller has sent which is wrong. It looks like some endpoints have missed implementing the eTags in general but we really should be doing it. We should probably add a ticket to go back through each of them and make sure they're all correct.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag

saminahbab: I have ammended the mongo query to retrieve the etag and add to the header

api/options.go
saminahbab: As discussed, and with a better understanding of `StatusCode(err)`, I have reworked the code to remove manual checks, lmk what you think

datastore/mongodb/error.go
saminahbab: as discussed, I have removed the additions to the interfaces, and just worked with what is there. New solution pushed. I will let you resolve if you are happy

api/api.go
```diff
@@ -88,9 +89,7 @@ func (api *API) enablePrivateEndpoints() {
 	r.Get("/filters/{id}/dimensions/{dimension}", api.getFilterDimension)
 	r.Post("/filters/{id}/dimensions", api.addFilterDimension)
 	r.Put("/filters/{id}/dimensions/{name}", api.updateFilterDimension)
-
-	r.Post("/filters/{id}/dimensions/{dimension}/options/{option}", api.addFilterDimensionOption)
```
ChrysmOre, line 92: Was this supposed to be deleted?

datastore/mongodb/options.go
```diff
@@ -0,0 +1,67 @@
...
+	"context"
+
+	"github.com/pkg/errors"
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
+)
+
+type optionQueryResult struct {
+	Options    []string `bson:"options"`
+	TotalCount int      `bson:"totalOptions"`
```
ChrysmOre, line 13: Should probably stick to snake_case for consistency

api/options.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	if pageLimit == 0 {
+		// define a reasonable default
+		// in light of bad input
+		// also slice not work with 0
+		pageLimit = 20
+	}
+
```
ChrysmOre, line 35: You didn't want to do any pre-emptive checks for the filter/dimension etc to return the other status codes? Not wanting to throw a spanner in the works after you've done all this work but I wonder if it wouldn't of been simpler to just retrieve the entire filter object from the store, perform the checks in order then return the options from the filter..

datastore/mongodb/options.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	var result []optionQueryResult
+
+	if err := c.conn.Collection(col.name).Aggregate(ctx, pipeline, &result); err != nil {
+		return nil, 0, "", errors.Wrap(err, "failed to get filter options")
+	}
+	if len(result) == 0 {
+		return nil, 0, "", &er{
+			err:      errors.Errorf("failed to find filter with ID (%s)", filterID),
```
ChrysmOre, line 53: We shouldn't be constructing error strings dynamically, use log.Data for that information.

datastore/mongodb/options.go
```diff
@@ -0,0 +1,67 @@
...
+			notFound: true,
+		}
+	}
+
+	options := result[0]
+
+	if options.TotalCount == -1 {
+		return nil, options.TotalCount, "", &er{
+			err: errors.Errorf("failed to find dimension with name (%s)", dimensionName),
```
ChrysmOre, line 62: Again here you should use log.Data

api/options.go
```diff
@@ -0,0 +1,99 @@
...
+			errors.Wrap(err, "Bad Request"),
+		)
+
+	}
+
+	if pageLimit == 0 {
+		// define a reasonable default
+		// in light of bad input
+		// also slice not work with 0
+		pageLimit = 20
```
ChrysmOre, line 33: Probably should use a const for this value, in fact I wouldn't be surprised if there is one already somewhere

api/options.go
saminahbab: I believe I've returned the right status codes as per the spec with one query, so It is 6 and half a dozen for me really

saminahbab: I will resolve for now if that is okay

ChrysmOre: Ok cool

ChrysmOre (approved):

PR #93, opened 2022-07-26
Fix/minor refactoring

ChrysmOre:
### What

Minor refactoring in api package for consistent URL param names

Upgraded dp-api-clients-go to latest version and use new fields

### How to review

Check changes make sense

### Who can review

Anyone


andre-urbani (approved):
LGTM

PR #94, opened 2022-07-26
Use correct dimensionID in filter

ChrysmOre:
### What

Fix to use return value from getFilterVariable as it presently attempts an additional lookup using the wrong value.

### How to review

Check change makes sense

### Who can review

Anyone


PauloPortugal (commented):
:+1: Thanks for fixing this.
I actually tested it, but clearly it wasn't thorough.

Please check the component tests.

PauloPortugal (approved):

j-s-rawat (approved):
LGTM

PR #95, opened 2022-07-28
Feature/5782 validation

saminahbab:
### What

Add validation step that ensures that only filters based on cantabular datasets are created 
Amend tests to include dataset information from the dataset api for create filter paths.

### How to review

run `make test-component` to ensure that the tests are passing

### Who can review

team b

### DIFF
--- a/features/create_filter.feature
+++ b/features/create_filter.feature
@@ -8,6 +8,7 @@ Feature: Filters Private Endpoints Not Enabled
       {
         "alerts": [],
         "collection_id": "dfb-38b11d6c4b69493a41028d10de503aabed3728828e17e64914832d91e1f493c6",
+        "is_based_on":{"@type": "flexible"},
         "dimensions": [
           {
             "label": "City",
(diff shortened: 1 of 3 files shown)


features/create_filter.feature
```diff
@@ -414,3 +468,43 @@ Feature: Filters Private Endpoints Not Enabled
     """
 
     And the HTTP status code should be "400"
+
+  Scenario: Creating a filter based on Cantabular blob
```
PauloPortugal, line 472: Maybe the `Scenario` description should be more explicit.

```suggestion
  Scenario:  Do not create a filter based on cantabular_table blob
```

features/create_filter.feature
```diff
@@ -414,3 +468,43 @@ Feature: Filters Private Endpoints Not Enabled
...
+            "Cardiff",
+            "London",
+            "Swansea"
+          ],
+          "is_area_type": true
+        }
+      ]
+    }
+    """
+    Then I should receive the following JSON response:
```
PauloPortugal, line 503: Prior to this check, I would also ensure that the response code is `400`

And the HTTP status code should be "400"
```suggestion
    Then the HTTP status code should be "400"
    And I should receive the following JSON response:
```

PauloPortugal (commented):
Looks good.
Small test readability improvements and please rebase develop into this branch.

PauloPortugal (dismissed):
:+1: Ace

PauloPortugal (approved):

PR #101, opened 2022-08-08
Feature/refactor component tests

PatrickConnollyONS:
### What

Refactoring components test for simplification and clarification:
- refactor test_main to create component once, register steps once, and reset component once after each scenario
- refactor Component to remove extraneous attributes and have a cleaner interface for RegisterSteps(), Reset() and Close()
- refactor steps.go to distribute the steps across the various Features (injectors) of the Component
- some test tidy ups/
- propose looking at refactoring/adding the distinction between authentication and authorisation

### How to review

Joshua to review the code 

### Who can review

Joshua Medley

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -29,7 +29,7 @@ type API struct {
 }
 
 // New creates and initialises a new API
-func New(ctx context.Context, cfg *config.Config, r chi.Router, idc *identity.Client, rsp responder, g generator, d datastore, ds datasetAPIClient, c cantabularClient, p kafka.IProducer) *API {
+func New(_ context.Context, cfg *config.Config, r chi.Router, idc *identity.Client, rsp responder, g generator, d datastore, ds datasetAPIClient, c cantabularClient, p kafka.IProducer) *API {
 	api := &API{
 		Router:         r,
 		respond:        rsp,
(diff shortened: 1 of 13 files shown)


features/steps/component.go
```diff
@@ -2,235 +2,101 @@ package steps
...
-	wg                *sync.WaitGroup
-	signals           chan os.Signal
-	ctx               context.Context
-	HTTPServer        *http.Server
-	store             service.Datastore
-	g                 service.Generator
-	shutdownInitiated bool
-	consumer          kafka.IConsumerGroup
-	waitEventTimeout  time.Duration
+	AuthServiceInjector *componenttest.AuthorizationFeature
```
ChrysmOre, line 32: I would really prefer we keep these named features rather than injectors, as it's what they're called in the packages and we're all used to that. I don't think you'll find anyone that finds it makes it considerably clearer what they do.

features/steps/cantabularfeature.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/cucumber/godog"
+)
+
+type CantabularFeature struct {
+	*mock.CantabularClient
+}
+
+func NewCantabularFeature() *CantabularFeature {
+	cf := &CantabularFeature{CantabularClient: &mock.CantabularClient{OptionsHappy: true}}
+	service.GetCantabularClient = func(cfg *config.Config) service.CantabularClient {
```
ChrysmOre, line 23: I feel like the New() should just be returning a new instance of whatever struct, not changing things in the service package. This should probably be done elsewhere

api/api.go
```diff
@@ -77,33 +77,44 @@ func (api *API) enablePublicEndpoints() {
 }
 
 func (api *API) enablePrivateEndpoints() {
-	r := chi.NewRouter()
+	// This is a hack to work around the issue whereby a chi.Router cannot set middleware after a route has been added
```
ChrysmOre, line 80: I'm not sure why this is necessary, if it's to enable something you've done in the component tests I'd rather not introduce hacks into our main service code to enable that.. It would be better if we could find a way so this wasn't necessary.

features/steps/component.go
PatrickConnollyONS: yeah I don't  have an issue with that

features/steps/cantabularfeature.go
PatrickConnollyONS: from what I can see the injection of the external dependencies is done either in the component (as at present) or the feature itself can do it?

api/api.go
PatrickConnollyONS: This one is a bit of a strange one. It's not directly related to component tests - as in the component tests will still work without the change I made - but I was looking at it when doing the component tests, since a component test can dynamically use the 'private endpoints are enabled/not enabled' (the reason the component tests will work is that the component is reset and re-initialised after every scenario, so the router will be re-initialised as well). But I made the change in an attempt to remove any prior knowledge of the state of the router when adding the private routes. Probably easier to have a conversation on this one, but I can remove it - I wanted to get some feedback on it first, and discuss what was there beforehand

features/steps/component.go
```diff
@@ -2,235 +2,100 @@ package steps
...
-func GetWorkingMongo(ctx context.Context, cfg *config.Config, g service.Generator) (service.Datastore, error) {
-	mongoClient, err := mongodb.NewClient(ctx, g, mongodb.Config{
-		MongoDriverConfig:       cfg.Mongo,
-		FilterFlexAPIURL:        cfg.BindAddr,
-		FiltersCollection:       cfg.FiltersCollection,
-		FilterOutputsCollection: cfg.FilterOutputsCollection,
-	})
-	if err != nil {
-		return nil, fmt.Errorf("failed to create new mongo mongoClient: %w", err)
+func setInitialiserMock(g service.Generator) {
```
ChrysmOre, line 93: Why remove this function from the component struct? Just means you have to pass extra arguments now

features/steps/component.go
```diff
@@ -2,235 +2,100 @@ package steps
...
-		waitEventTimeout: WaitEventTimeout,
-	}, nil
-}
-
-// Init initialises the server, the mocks and waits for the dependencies to be ready
-func (c *Component) Init() (http.Handler, error) {
-	c.signals = make(chan os.Signal, 1)
-	signal.Notify(c.signals, os.Interrupt, syscall.SIGTERM)
+	cfg.ZebedeeURL = component.AuthFeature.FakeAuthService.ResolveURL("")
+	log.Info(context.Background(), "config used by component tests", log.Data{"cfg": cfg})
```
ChrysmOre, line 58: Could we pass the relevant context to use here rather than background()?

PatrickConnollyONS: I don't think we have a relevant context at this point? To be honest I'd prefer to get rid of the log altogether - it may be okay while debugging the tests but the tests should not need to be printing/logging - would you be okay to take it out?

features/steps/component.go
PatrickConnollyONS: I have added this function back as a method on component, which combines what we talked about yesterday re: features setting their service injection points. The reason 'g' is passed as a parameter is to avoid cluttering the component object with unnecessary state - after initialisation g is never accessed by the component object again.

features/steps/component.go
```diff
@@ -2,235 +2,102 @@ package steps
...
-	service.GetCantabularClient = func(_ *config.Config) service.CantabularClient {
-		return c.CantabularClient
-	}
-
-	service.GetMongoDB = func(ctx context.Context, cfg *config.Config, g service.Generator) (service.Datastore, error) {
-		return c.store, nil
+// Router initialises the service, returning the service's (server) router for tests
+// This delayed initialisation is needed to ensure that any changes to the router (or the service in general)
+// as a result of test setup, are picked up
+func (c *Component) Router() (http.Handler, error) {
```
ChrysmOre, line 68: Is this function used anywhere?

features/steps/component.go
```diff
@@ -2,235 +2,102 @@ package steps
...
-// Init initialises the server, the mocks and waits for the dependencies to be ready
-func (c *Component) Init() (http.Handler, error) {
-	c.signals = make(chan os.Signal, 1)
-	signal.Notify(c.signals, os.Interrupt, syscall.SIGTERM)
-
-	log.Info(c.ctx, "config used by component tests", log.Data{"cfg": c.cfg})
+	cfg.ZebedeeURL = component.AuthFeature.FakeAuthService.ResolveURL("")
 
-	c.cfg.DatasetAPIURL = c.DatasetAPI.ResolveURL("")
+	component.setInitialiserMock(g)
```
ChrysmOre, line 58: Any reason why the generator can't be part of the component struct (and not having to pass it round like this)?

features/steps/component.go
PatrickConnollyONS: yes this is the function that is passed to the APIFeature (in func NewComponent():
component.APIFeature = componenttest.NewAPIFeature(component.Router)
and the feature uses it to retrieve the router just before it sends a request to one of the router endpoints. This is the function that ensures the component (the service) is initialised with the correct test setup for the test that the APIFeature is about to call (i.e. in the 'When I Get/Put/Post...' step)

features/steps/component.go
PatrickConnollyONS: One of the goals I had was to try to reduce/minimise the sometimes large amount of 'state' that was being stored in the component object - really all I wanted to see in the object was the service (the actual component under test) and the Features that are used as the service's dependencies (or non dependency essential features such as the API/Auth/ErrorFeature). I removed the generator because once it has been set in setInitialiserMock() is is never used again, and storing it just seemed redundant (and maybe, as in my case, can seem confusing as to why it's there)

ChrysmOre (approved):

PR #105, opened 2022-08-31
Feature/5826 multivariate dimension validation

saminahbab:
### What

[Trello Ticket Here](https://trello.com/c/DdCjzC84/5826-add-validation-in-dp-cantabular-filter-flex-api)

Added notion of `cantabular_flexible_table` to the creation of filters and validation 
Added tests to show that multiple non geography dimensions can be added to a filter 
Added tests to show that a multivariable filter can be created 
### How to review

Run make test and make test-component 
Inspect the tests to ensure that the right inputs and outputs are given in the tests for multivariate dimension filters

### Who can review

team b

### DIFF
--- a/features/create_filter.feature
+++ b/features/create_filter.feature
@@ -469,6 +469,51 @@ Feature: Filters Private Endpoints Not Enabled
 
     And the HTTP status code should be "400"
 
+    Scenario: Creating a new invalid request (duplicate dimensions)
+
+    When I POST "/filters"
+    """
+    {
+      "dataset":{
+          "id":      "cantabular-example-1",
+          "edition": "2021",
+          "version": 1
+       },
+      "population_type": "Example",
+      "dimensions": [
+        {
+          "name": "siblings",
+          "options": [
+            "0-3",
+            "4-7",
+            "7+"
+          ],
+          "is_area_type": false
+        },
+        {
+          "name": "siblings",
+          "options": [
+            "0-3",
+            "4-7",
+            "7+"
+          ],
+          "is_area_type": false
+        }
+      ]
+    }
+    """
+
+    Then I should receive the following JSON response:
+    """
+    {
+      "errors": [
+        "failed to validate request dimensions: duplicate dimensions chosen: siblings"
+      ]
+    }
+    """
+
+    And the HTTP status code should be "404"  
+
   Scenario: Do not create a filter based on cantabular blob
     When I POST "/filters"
     """


api/filters.go
```diff
@@ -19,6 +19,7 @@ import (
 
 const (
 	flexible        = "flexible"
+	multivariate    = "cantabular_multivariate_table"
```
ChrysmOre, line 22: We should keep the filter types consistent, `multivariate` like `flexible`

api/filters.go
```diff
@@ -93,7 +94,8 @@ func (api *API) createFilter(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if err := api.isValidDatasetDimensions(ctx, v, req.Dimensions, req.PopulationType); err != nil {
+	// update here for the first
+	if err := api.isValidDatasetDimensions(ctx, v, req.Dimensions, req.PopulationType, v.IsBasedOn.Type); err != nil {
```
ChrysmOre, line 98: You seem to be passing two different 'types' here, dataset type in one place and filter type another, I don't think they don't hold the same set of values.. where a filter would be 'flexible'  a dataset would be of type 'cantabular_flexible_table'.

api/filters.go
```diff
@@ -122,6 +124,10 @@ func (api *API) createFilter(w http.ResponseWriter, r *http.Request) {
 		DisclosureControl: nil, // populate for these fields yet
 	}
 
+	if v.IsBasedOn.Type == "cantabular_multivariate_table" {
```
ChrysmOre, line 127: You're using a const and a non-const for the same value.. Above I mentioned the filter type should probably be changed. You should also add a `cantabularMultivariateTable` const and use that everywhere you need too.

api/filters.go
```diff
@@ -334,8 +340,8 @@ func (api *API) putFilter(w http.ResponseWriter, r *http.Request) {
 	api.respond.JSON(ctx, w, http.StatusOK, resp)
 }
 
-func (api *API) isValidDatasetDimensions(ctx context.Context, v dataset.Version, d []model.Dimension, pType string) error {
-	dimIDs, err := api.validateDimensions(d, v.Dimensions)
+func (api *API) isValidDatasetDimensions(ctx context.Context, v dataset.Version, d []model.Dimension, pType string, cantabularType string) error {
```
ChrysmOre, line 343: You're calling this `cantabularType` here and `datasetType` elsewhere. Personally Ithink `Type` makes more sense, and is more obvious what it is. But do keep it consistent

api/filters.go
```diff
@@ -402,10 +408,18 @@ func (api *API) validateDimensions(filterDims []model.Dimension, dims []dataset.
...
+
 	}
 	var incorrect []string
 	for _, fd := range filterDims {
 		if _, ok := dimensions[fd.Name]; !ok {
+			// if this is a cantabular multivariate table, then you
+			// should be free to create whatever dimension that you
+			// want to add, provided that they exist. which is checked
+			// in dimension option validation.
+			if datasetType == "cantabular_multivariate_table" {
```
ChrysmOre, line 420: IIRC the dimension options check only performs the test if there are specific options chosen. If this is left blank(where users would want all options) then no check will be done. Another solution is likely needed

api/filters.go
```diff
@@ -402,10 +408,18 @@ func (api *API) validateDimensions(filterDims []model.Dimension, dims []dataset.
 	dimensions := make(map[string]string)
 	for _, d := range dims {
 		dimensions[d.Name] = d.ID
+
```
ChrysmOre, line 411: Blank line

api/filters_test.go
```diff
@@ -81,7 +81,7 @@ func TestValidateDimensions(t *testing.T) {
 				"bar": "bar01",
 			}
 
-			got, err := api.validateDimensions(filterDims, existingDims)
+			got, err := api.validateDimensions(filterDims, existingDims, "flexible")
```
ChrysmOre, line 84: is there a const we can use here?

api/filters.go
ChrysmOre: Also, this check could be moved out of the loop just to return straight away here as nothing else will be done

PR #109, opened 2022-09-05
Feature/5826 validation

saminahbab:
### What

[Trello Ticket Here](https://trello.com/c/DdCjzC84/5826-add-validation-in-dp-cantabular-filter-flex-api)

Add validation for multivariate filters. 
separate out flexible validation. 
Add Checks for cantabular dimension existence when multivariate. 
Ensure that component tests are added to show multivariate ability. 

### How to review

Check the component tests to ensure that the test cases are covered. 

### Who can review

Team B can review


api/filter_dimensions.go
```diff
@@ -88,9 +90,25 @@ func (api *API) addFilterDimension(w http.ResponseWriter, r *http.Request) {
...
+	} else if filter.Type == multivariate {
+
+		multivariateDim, err := api.isValidMultivariateDimensions(ctx, toValidate, filter.PopulationType)
+		if err != nil {
+			api.respond.Error(ctx, w, statusCode(err), err)
+		}
+
+		finalDim = multivariateDim[0]
+
 	}
```
PauloPortugal, line 113: Probably encapsulate this into method that can be reused on this `api` package, both on the `api/filter.go` and on the `api/filter_dimensions.go` files.

api/filter_dimensions.go
```diff
@@ -88,9 +90,26 @@ func (api *API) addFilterDimension(w http.ResponseWriter, r *http.Request) {
...
 
-	if err := api.isValidDatasetDimensions(ctx, v, []model.Dimension{req.Dimension}, filter.PopulationType); err != nil {
-		api.respond.Error(ctx, w, statusCode(err), err)
-		return
+	toValidate := []model.Dimension{req.Dimension}
+
+	if filter.Type == flexible {
+		err := api.isValidDatasetDimensions(ctx, v, toValidate, filter.PopulationType)
+		if err != nil {
+			api.respond.Error(ctx, w, statusCode(err), err)
```
ChrysmOre: Missed a return here

api/filter_dimensions.go
```diff
@@ -88,9 +90,26 @@ func (api *API) addFilterDimension(w http.ResponseWriter, r *http.Request) {
...
+		}
+
```
ChrysmOre: Lots of unnecessary blank lines at the beginning and end of scopes

api/filter_dimensions.go
```diff
@@ -88,9 +90,26 @@ func (api *API) addFilterDimension(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if err := api.isValidDatasetDimensions(ctx, v, []model.Dimension{req.Dimension}, filter.PopulationType); err != nil {
-		api.respond.Error(ctx, w, statusCode(err), err)
-		return
+	toValidate := []model.Dimension{req.Dimension}
+
+	if filter.Type == flexible {
+		err := api.isValidDatasetDimensions(ctx, v, toValidate, filter.PopulationType)
```
ChrysmOre: this could be one line `if err := api.isValid..; err != nil{`

api/filter_dimensions.go
saminahbab: Hello Paulo, please find encapsulation in the latest commit, alongside ammended tests to marry up the check with expectations. I will leave  you to resolve the comment

api/filters.go
```diff
@@ -93,9 +99,28 @@ func (api *API) createFilter(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if err := api.isValidDatasetDimensions(ctx, v, req.Dimensions, req.PopulationType); err != nil {
-		api.respond.Error(ctx, w, statusCode(err), err)
-		return
+	if v.IsBasedOn.Type == cantabularFlexible {
+
```
ChrysmOre: lots of extra whitespace

api/filters.go
```diff
@@ -353,6 +411,8 @@ func (api *API) isValidDatasetDimensions(ctx context.Context, v dataset.Version,
 // getCantabularDimension checks that dimension exists in Cantabular by searching for it.
 // If the dimension doesn't exist, or couldn't be retrieved, an error is returned.
 func (api *API) getCantabularDimension(ctx context.Context, filterID, dimensionName string) (*gql.Node, error) {
+	// No idea why the filter is being retrieved here?
```
ChrysmOre, line 400: using filter.PopulationType just below in the search request

api/filters.go
```diff
@@ -381,6 +441,29 @@ func (api *API) getCantabularDimension(ctx context.Context, filterID, dimensionN
 	return &foundDimensions.Dataset.Variables.Search.Edges[0].Node, nil
 }
 
+// getCantabularDimension checks that dimension exists in Cantabular by searching for it.
+// If the dimension doesn't exist, or couldn't be retrieved, an error is returned.
+func (api *API) getCantabularDimensionNoFilterID(ctx context.Context, popType, dimensionName string) (*gql.Node, error) {
+
```
ChrysmOre, line 433: whitespace

api/filters.go
```diff
@@ -381,6 +427,29 @@ func (api *API) getCantabularDimension(ctx context.Context, filterID, dimensionN
 	return &foundDimensions.Dataset.Variables.Search.Edges[0].Node, nil
 }
 
+// getCantabularDimension checks that dimension exists in Cantabular by searching for it.
+// If the dimension doesn't exist, or couldn't be retrieved, an error is returned.
+func (api *API) getCantabularDimensionNoFilterID(ctx context.Context, popType, dimensionName string) (*gql.Node, error) {
```
j-s-rawat, line 432: Can we please remove getCantabularDimensionNoFilterID and update  getCantabularDimension where population_type is passed as parameter rather than filter_id

ChrysmOre (commented):
Some comments r.e formatting.

A bit worrying a missing return wasn't picked up by tests or otherwise.

I suspect the existing code around getCantabularDimension could be refactored take the populationType directly rather than the filterID, removing the need for two separate functions. I may be wrong but would probably be preferable if possible

saminahbab:
> Some comments r.e formatting.
> 
> A bit worrying a missing return wasn't picked up by tests or otherwise.
> 
> I suspect the existing code around getCantabularDimension could be refactored take the populationType directly rather than the filterID, removing the need for two separate functions. I may be wrong but would probably be preferable if possible

Yeah I actually put a test in before this comment about putting in missing dimensions, and caught it before I saw the comment apologies for that, will work on the refactor on the getCantabularDimension right now

api/filters.go
saminahbab: done

api/filter_dimensions.go
```diff
@@ -87,12 +83,12 @@ func (api *API) addFilterDimension(w http.ResponseWriter, r *http.Request) {
 		)
 		return
 	}
-
-	if err := api.isValidDatasetDimensions(ctx, v, []model.Dimension{req.Dimension}, filter.PopulationType); err != nil {
+	dimensions, _, err := api.ValidateCantabularFilter(v, []model.Dimension{req.Dimension}, filter.PopulationType)
```
j-s-rawat, line 86: To me the name of function is a bit misleading as it is not validating the filter, rather is running validations and actually populating/returning dimensions.

api/filter_dimensions.go
```diff
@@ -87,12 +83,12 @@ func (api *API) addFilterDimension(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
 		api.respond.Error(ctx, w, statusCode(err), err)
 		return
 	}
-
+	finalDim := dimensions[0]
```
j-s-rawat, line 91: can we please add some comment to help understand 'final dimension'

api/filter_dimensions.go
saminahbab: true, will change

api/filter_dimensions.go
saminahbab: I just didn't want to reference dimensions[0] all the time, not sure if it is worth a comment

j-s-rawat (approved):
LGTM

PR #114, opened 2022-09-15
Release/1.7.0

ChrysmOre:
### What

Release 1.7.0

- Always require IsAreaType to be provided with dimension requests
- 
### How to review

Check correct commits present

### Who can review

Anyone who hasn't contributed

### DIFF
--- a/api/contract.go
+++ b/api/contract.go
@@ -45,6 +45,10 @@ func (r *createFilterRequest) Valid() error {
 			return fmt.Errorf("missing field: [dimension[%d].name]", i)
 		}
 
+		if d.IsAreaType == nil {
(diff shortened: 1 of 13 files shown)


api/contract.go
```diff
@@ -45,6 +45,10 @@ func (r *createFilterRequest) Valid() error {
 			return fmt.Errorf("missing field: [dimension[%d].name]", i)
 		}
 
+		if d.IsAreaType == nil {
```
franmoore05, line 48: This has caused issues with the frontend in sandbox so has had to be taken out

franmoore05 (commented):
There is code in here which has caused our front end to break which Jitendar is going to revert so we can continue using the service.

PR #115, opened 2022-09-16
Temporary change to fix when IsAreaType is passes as nil from florence

j-s-rawat:
### What
Removed the check that is crashing the creation of filter from florence

### How to review

All tests are green

### Who can review
Anyone


PauloPortugal (approved):

PauloPortugal:
I imagine you have tested this locally.
It would be good to hear from @franmoore05 about this or @ChrysmOre as he was the last person to change the logic.

PR #118, opened 2022-09-21
Feature/geography endpoints

PatrickConnollyONS:
### What

Add the 'geography' endpoint (/datasets/{dataset_id}/editions/{edition}/versions/{version}/json)
with associated unit and component tests

### How to review

Review code and ensure all tests pass

### Who can review

Anyone familiar with cantabular datasets

### DIFF
--- a/api/dataset.go
+++ b/api/dataset.go
@@ -34,7 +34,6 @@ func (api *API) getDatasetJSONHandler(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 
 	params, err := api.getDatasetParams(ctx, r)
-
 	if err != nil {
 		api.respond.Error(
 			ctx,
@@ -52,7 +51,6 @@ func (api *API) getDatasetJSONHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	response, err := api.getDatasetJSON(ctx, r, params)
-
 	if err != nil {
 		api.respond.Error(
 			ctx,
@@ -75,8 +73,7 @@ func (api *API) getDatasetJSON(ctx context.Context, r *http.Request, params *dat
 	}
 
 	if r.URL.Query().Get("geography") != "" {
-		filters, err := api.getGeographyFilters(ctx, r, params)
-
+		filters, err := api.getGeographyFilters(r, params)
 		if err != nil {
 			return nil, errors.Wrap(err, "getGeographyFilters failed")
 		}
@@ -97,7 +94,7 @@ func (api *API) getDatasetJSON(ctx context.Context, r *http.Request, params *dat
 	return response, nil
 }
 
-func (api *API) getGeographyFilters(_ context.Context, r *http.Request, params *datasetParams) ([]cantabular.Filter, error) {
+func (api *API) getGeographyFilters(r *http.Request, params *datasetParams) ([]cantabular.Filter, error) {
 	geographyQuery := strings.Split(r.URL.Query().Get("geography"), ",")
 
 	if len(geographyQuery) < 2 {
@@ -108,8 +105,8 @@ func (api *API) getGeographyFilters(_ context.Context, r *http.Request, params *
 	geographyOptions := geographyQuery[1:]
 
 	foundGeography := false
-	for _, datasetGeography := range params.geoDimensions {
-		if strings.EqualFold(datasetGeography, geography) {
+	for _, d := range params.geoDimensions {
+		if strings.EqualFold(d, geography) {
 			foundGeography = true
 			break
 		}
@@ -120,10 +117,10 @@ func (api *API) getGeographyFilters(_ context.Context, r *http.Request, params *
 	}
 
 	var geographyCodes []string
-	for _, geographyOption := range geographyOptions {
-		opt, ok := params.options[geography][geographyOption]
+	for _, o := range geographyOptions {
+		opt, ok := params.options[geography][o]
 		if !ok {
-			return nil, errors.Errorf("unable to validate geography option %s", geographyOption)
+			return nil, errors.Errorf("unable to validate geography option %s", o)
 		}
 		geographyCodes = append(geographyCodes, opt.Option)
 	}
@@ -134,8 +131,8 @@ func (api *API) getGeographyFilters(_ context.Context, r *http.Request, params *
 	}
 
 	foundDimension := false
-	for _, datasetDimension := range params.datasetDimensions {
-		if datasetDimension == dimension {
+	for _, d := range params.datasetDimensions {
+		if d == dimension {
 			foundDimension = true
 			break
 		}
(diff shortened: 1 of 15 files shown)


api/dataset.go
```diff
@@ -9,38 +9,39 @@ import (
...
-	if r.URL.Query().Get("geography") == "" {
-		api.getDefaultDatasetJSON(ctx, w, r)
-	} else {
-		api.getGeographyDatsetJSON(ctx, r)
-	}
-}
-
-func (api *API) getDefaultDatasetJSON(ctx context.Context, w http.ResponseWriter, r *http.Request) {
 	params, err := api.getDatasetParams(ctx, r)
+
```
ChrysmOre, line 37: No need to leave a blank line between checking errors

api/dataset.go
```diff
@@ -50,63 +51,117 @@ func (api *API) getDefaultDatasetJSON(ctx context.Context, w http.ResponseWriter
 		"version": params.version,
 	}
 
-	geoDimensions, err := api.getGeographyTypes(ctx, params.basedOn)
+	response, err := api.getDatasetJSON(ctx, r, params)
+
```
ChrysmOre, line 55: another blank line

api/dataset.go
```diff
@@ -50,63 +51,117 @@ func (api *API) getDefaultDatasetJSON(ctx context.Context, w http.ResponseWriter
...
+func (api *API) getDatasetJSON(ctx context.Context, r *http.Request, params *datasetParams) (*GetDatasetJSONResponse, error) {
 	datasetRequest := cantabular.StaticDatasetQueryRequest{
 		Dataset:   params.basedOn,
-		Variables: dimensions,
+		Variables: params.sortedDimensions,
+	}
+
+	if r.URL.Query().Get("geography") != "" {
+		filters, err := api.getGeographyFilters(ctx, r, params)
+
```
ChrysmOre, line 79: blank line

api/dataset.go
```diff
@@ -50,63 +51,117 @@ func (api *API) getDefaultDatasetJSON(ctx context.Context, w http.ResponseWriter
...
+		return nil, errors.Wrap(err, "failed to generate response")
 	}
 
-	api.respond.JSON(ctx, w, http.StatusOK, response)
+	return response, nil
 }
 
-func (api *API) getGeographyDatsetJSON(ctx context.Context, r *http.Request) (*getDatasetJSONResponse, error) {
-	return &getDatasetJSONResponse{}, nil
+func (api *API) getGeographyFilters(_ context.Context, r *http.Request, params *datasetParams) ([]cantabular.Filter, error) {
```
ChrysmOre, line 100: If ctx is blank is it necessary to have it at all?

api/dataset.go
```diff
@@ -50,63 +51,117 @@ func (api *API) getDefaultDatasetJSON(ctx context.Context, w http.ResponseWriter
...
+
+	if len(geographyQuery) < 2 {
+		return nil, errors.New("unable to locate geography")
+	}
+
+	geography := geographyQuery[0]
+	geographyOptions := geographyQuery[1:]
+
+	foundGeography := false
+	for _, datasetGeography := range params.geoDimensions {
```
ChrysmOre, line 111: It's preferable to use shorter variable names where possible, especially in places where they are short lived/have limited scope as in a for loop. Something like `for _, d := range..` would be more typical

api/dataset.go
```diff
@@ -180,23 +233,20 @@ func (api *API) getDatasetParams(ctx context.Context, r *http.Request) (*dataset
...
+		return nil, errors.Wrap(err, "failed to get version")
 	}
 
-	params.datasetLink = datasetItem.Links.Self
-	params.basedOn = datasetItem.IsBasedOn.ID
+	params.datasetLink = versionItem.Links.Dataset
+	params.basedOn = versionItem.IsBasedOn.ID
 
-	if datasetItem.DatasetDetails.Type != "cantabular_flexible_table" {
+	if versionItem.IsBasedOn.Type != "cantabular_flexible_table" {
```
ChrysmOre, line 246: in filters.go there are consts for the various table types that can be used

features/steps/steps.go
```diff
@@ -218,3 +236,230 @@ func (c *Component) theFollowingExportStartEventsAreProduced(events *godog.Table
...
+		}
+
+		dimOpts = append(dimOpts, opts...)
+	}
+
+	return dimOpts
+}
+
+func (c *Component) theDatasetJSONResult(expected *godog.DocString) error {
+	var got, expt api.GetDatasetJSONResponse
```
ChrysmOre, line 404: What's wrong with the built in `I should receive the following JSON response` step?

features/steps/steps.go
```diff
@@ -59,6 +68,15 @@ func (c *Component) RegisterSteps(ctx *godog.ScenarioContext) {
 	ctx.Step(`^the following Export Start events are produced:$`,
 		c.theFollowingExportStartEventsAreProduced,
 	)
+	ctx.Step(`^the following recipe is used to create a dataset based on the given cantabular dataset:$`,
```
ChrysmOre, line 71: Is this the right way to be testing this? It's quite a large layer of abstraction in the test code.. Wouldn't it be better to explicitly be laying out what the service actually interacts with? I.E. the available dataset, metadata, dimensions etc that this step is comprised of?

features/steps/cantabularfeature.go
```diff
@@ -3,34 +3,66 @@ package steps
...
 	"github.com/ONSdigital/dp-cantabular-filter-flex-api/features/mock"
 	"github.com/ONSdigital/dp-cantabular-filter-flex-api/service"
+	dphttp "github.com/ONSdigital/dp-net/v2/http"
 
 	"github.com/cucumber/godog"
 )
 
 type CantabularFeature struct {
-	*mock.CantabularClient
+	cantabularMock   *mock.CantabularClient
```
ChrysmOre, line 22: I think cantabularClient was a clearer name to differentiate from cantabularServer

ChrysmOre (commented):
A few comments, mostly R.E formatting etc

api/dataset.go
PatrickConnollyONS: I removed it - probably cleaner

features/steps/steps.go
PatrickConnollyONS: This was to provide some extras over what the `I should receive the following JSON response` does. Specifically because the tests use a fake cantabular server, I want to ignore the hostname embedded in certain urls, and this function does that.
This concern might disappear depending on what we do with the next issue below 👇

features/steps/steps.go
PatrickConnollyONS: looking at it both ways might give the answer

features/steps/steps.go
PatrickConnollyONS: I have changed this to use the standard step: `I should receive the following JSON response`

features/steps/steps.go
PatrickConnollyONS: I have added a get-dataset-json-alternate.feature to show the 2 different approaches. The latter is a lot more verbose but see what you think

ChrysmOre (approved):

ChrysmOre (approved):

PR #122, opened 2022-09-23
patch: remove filter output locking

saminahbab:
### What

Comment out the locking when updating filter outputs. 
This is to experiment whether the 300 concurrent user test can pass. 
Tests were left as is, because it is unclear whether this locking will stay in the future.

### How to review

run `make test` and `make test-component` to see no changes 
inspect commented code

### Who can review

team b


PauloPortugal (approved):
Looks good to me.

PR #140, opened 2022-10-18
Validate all hydrate all greography dimensions from Cantabular rather…

ChrysmOre:
… than version

### What

Allow users to create filters with Geography dimensions other than the one found in the Version document for all filter types.

This is achieved by sending the Geography dimension to Cantabular for validation and hydration as we do for multivariate filters.

### How to review

Check changes make sense

### Who can review

Anyone


api/validation.go
```diff
@@ -19,11 +19,23 @@ func (api *API) validateAndHydrateDimensions(v dataset.Version, dims []model.Dim
 	ctx := context.Background()
 
 	if v.IsBasedOn.Type == cantabularFlexibleTable {
+		var geodim model.Dimension
+		for _, d := range dims {
+			if d.IsAreaType != nil && *d.IsAreaType {
+				geodims, err := api.getCantabularDimensions(ctx, []model.Dimension{d}, pType)
+				if err != nil {
```
redhug1, line 26: What happens if this:
https://github.com/ONSdigital/dp-cantabular-filter-flex-api/blob/1e9d1fd66d750a8d05f3520ab28061321fb99b74/api/validation.go#L59

is passed an empty array ?
Will it return an empty array that has nothing in it, and could that then make line 29 below blow up ?

ChrysmOre: At the moment it would be impossible as the api calls would fail initial validation but I've added some extra checks to make sure

redhug1 (changes requested):
Please update line 9 of build.yml to be:
    tag: 1.19.2

and similarly for unit.yml

redhug1 (approved):

PR #150, opened 2022-10-31
Feature/amendments to json endpoint

franmoore05:
### What

Amended /json endpoint so a geography can now be fully flexed.  Added a new function to do this as the multivariates work in future may need the previous one created so we can keep it.

### How to review

Ensure the tests pass and the code makes sense.

### Who can review

Anyone.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -118,5 +118,7 @@ func (api *API) enablePrivateEndpoints() {
 	r.Put("/filter-outputs/{id}", api.putFilterOutput)
 	r.Post("/filter-outputs/{id}/events", api.addFilterOutputEvent)
 
+	r.Get("/datasets/{dataset_id}/editions/{edition}/versions/{version}/json", api.getDatasetJSONHandler)
+
 	api.Router.Mount("/", r)
 }
--- a/api/dataset.go
+++ b/api/dataset.go
@@ -75,8 +75,8 @@ func (api *API) getDatasetJSON(ctx context.Context, r *http.Request, params *dat
 
 	fmt.Println(params.sortedDimensions)
 
-	if r.URL.Query().Get("geography") != "" {
-		geographyQuery := strings.Split(r.URL.Query().Get("geography"), ",")
+	if r.URL.Query().Get("area-type") != "" {
+		geographyQuery := strings.Split(r.URL.Query().Get("area-type"), ",")
 
 		success, err := api.validateGeography(ctx, r, params)
 		if err != nil {
@@ -110,7 +110,7 @@ func (api *API) getDatasetJSON(ctx context.Context, r *http.Request, params *dat
 }
 
 func (api *API) validateGeography(ctx context.Context, r *http.Request, params *datasetParams) (bool, error) {
-	geographyQuery := strings.Split(r.URL.Query().Get("geography"), ",")
+	geographyQuery := strings.Split(r.URL.Query().Get("area-type"), ",")
 
 	fmt.Println(geographyQuery)
 	if len(geographyQuery) < 1 {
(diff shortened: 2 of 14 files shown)


api/dataset.go
```diff
@@ -94,8 +94,56 @@ func (api *API) getDatasetJSON(ctx context.Context, r *http.Request, params *dat
 	return response, nil
 }
 
+func (api *API) validateGeography(ctx context.Context, r *http.Request, params *datasetParams, datasetRequest cantabular.StaticDatasetQueryRequest) ([]string, *[]cantabular.Filter, error) {
```
ChrysmOre, line 97: Never any need to make a pointer to a slice as a slice essentially acts like a pointer in itself (including being able to equal nil)

api/dataset.go
```diff
@@ -94,8 +94,56 @@ func (api *API) getDatasetJSON(ctx context.Context, r *http.Request, params *dat
...
+		// Check the area code is in the area type
+		for _, a := range geographyOptions {
+			cReq := cantabular.GetAreaRequest{
+				Dataset:  params.basedOn,
+				Variable: geography,
+				Category: a,
+			}
+
+			_, err := api.ctblr.GetArea(ctx, cReq)
+
```
ChrysmOre, line 128: better not to leave blank lines before checking errors

api/dataset.go
```diff
@@ -94,8 +94,56 @@ func (api *API) getDatasetJSON(ctx context.Context, r *http.Request, params *dat
...
+	if len(geographyOptions) > 0 {
+		// Check the area code is in the area type
+		for _, a := range geographyOptions {
+			cReq := cantabular.GetAreaRequest{
+				Dataset:  params.basedOn,
+				Variable: geography,
+				Category: a,
+			}
+
+			_, err := api.ctblr.GetArea(ctx, cReq)
```
ChrysmOre, line 127: Isn't there a `cantabular.GetAreas` function that could query all at once?

api/dataset.go
```diff
@@ -156,22 +204,23 @@ func (api *API) getGeographyFilters(r *http.Request, params *datasetParams) ([]c
...
 }
 
 func (api *API) toGetDatasetJsonResponse(params *datasetParams, query *cantabular.StaticDatasetQuery) (*GetDatasetJSONResponse, error) {
 	var dimensions []DatasetJSONDimension
 
 	for _, dimension := range query.Dataset.Table.Dimensions {
-		if _, ok := params.options[dimension.Variable.Name]; !ok {
-			return nil, errors.New("dimension mismatch")
-		}
+		// if _, ok := params.options[dimension.Variable.Name]; !ok {
```
ChrysmOre, line 214: Meant to be still commented out or removed?

api/api.go
```diff
@@ -108,5 +108,7 @@ func (api *API) enablePrivateEndpoints() {
 	r.Put("/filter-outputs/{id}", api.putFilterOutput)
 	r.Post("/filter-outputs/{id}/events", api.addFilterOutputEvent)
 
+	r.Get("/datasets/{dataset_id}/editions/{edition}/versions/{version}/json", api.getDatasetJSONHandler)
```
ChrysmOre, line 111: If the endpoint is now meant to be available to publishing then some secrets need to be updated for dp-api-router

ChrysmOre (commented):
Some comments

api/dataset.go
j-s-rawat: This is a test to validate that all geography options are valid. There is a possibility of getting all the areas i.e. GetAreas, and then check if geography options are valid in  list returned by GetAreas, but GetAreas is based on pagination and will have to implement pagination logic to retrieve all the list. 
So unless geography options is a big list and cantabular.GetArea is inefficient, there  is no harm in the way it is currently implemented.

api/api.go
j-s-rawat: Create a [PR](https://github.com/ONSdigital/dp-configs/pull/1131) to address this issue

thomasridd (dismissed):
This is a priority ticket and all PR comments have been addressed. I'm approving it to open up the opportunity to merge

ChrysmOre (approved):

PR #151, opened 2022-11-03
add lowest_geography to dataset model

andre-urbani:
### What

Add `lowest_geogrpahy` to `Dataset` model
Trello - https://trello.com/c/lmWyoTS6/5923-add-lowestgeography-to-filterdataset-model-in-filter-api-and-filter-flex-api

### How to review

Confirm changes match those requested in Trello ticket

### Who can review

Anyone

### DIFF
--- a/api/contract.go
+++ b/api/contract.go
@@ -28,8 +28,8 @@ func (r *createFilterRequest) Valid() error {
 		return errors.New("missing field: dataset")
 	}
 
-	if r.Dataset.ID == "" || r.Dataset.Edition == "" || r.Dataset.Version == 0 {
-		return errors.New("missing field: [dataset.id | dataset.edition | dataset.version]")
+	if r.Dataset.ID == "" || r.Dataset.Edition == "" || r.Dataset.Version == 0 || r.Dataset.LowestGeography == "" {
+		return errors.New("missing field: [dataset.id | dataset.edition | dataset.version | dataset.lowest_geography]")
 	}
 
 	if r.PopulationType == "" {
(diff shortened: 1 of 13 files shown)


api/contract.go
```diff
@@ -28,8 +28,8 @@ func (r *createFilterRequest) Valid() error {
 		return errors.New("missing field: dataset")
 	}
 
-	if r.Dataset.ID == "" || r.Dataset.Edition == "" || r.Dataset.Version == 0 {
-		return errors.New("missing field: [dataset.id | dataset.edition | dataset.version]")
+	if r.Dataset.ID == "" || r.Dataset.Edition == "" || r.Dataset.Version == 0 || r.Dataset.LowestGeography == "" {
+		return errors.New("missing field: [dataset.id | dataset.edition | dataset.version | dataset.lowest_geography]")
```
ChrysmOre, line 32: I don't think this is right.. we shouldn't need to rely on the user to specify the lowest_geography, it should be saved to the dataset for us to retrieve when we call GetVersion. It's been added to the instance in dataset-api and dp-api-clients-go in my PRs.

api/contract.go
```diff
@@ -29,7 +29,7 @@ func (r *createFilterRequest) Valid() error {
 	}
 
 	if r.Dataset.ID == "" || r.Dataset.Edition == "" || r.Dataset.Version == 0 {
-		return errors.New("missing field: [dataset.id | dataset.edition | dataset.version]")
+		return errors.New("missing field: [dataset.id | dataset.edition | dataset.version")
```
ChrysmOre, line 32: looks like you accidentally deleted the ']'

ChrysmOre (approved):

PR #152, opened 2022-11-03
patch: remove dimension validation

saminahbab:
### What

Removed the validation for single dimension filters, allowing them to pass through. 
Ammend tests to represent new validation rules 

### How to review

Check tests 

### Who can review

team b


api/contract_test.go
```diff
@@ -73,7 +73,7 @@ func TestCreateFiltersRequestValid(t *testing.T) {
 			r.Dimensions = r.Dimensions[:1]
 			Convey("When Valid() is called", func() {
 				err := r.Valid()
-				So(err, ShouldNotBeNil)
+				So(err, ShouldBeNil)
```
j-s-rawat, line 76: This test no longer make much sense. Will it be better to update the test where a filter with 0 dimensions return an error, else we can remove the test

saminahbab: I just removed it

datastore/mongodb/filter_dimensions.go
```diff
@@ -342,7 +342,7 @@ func (c *Client) DeleteFilterDimension(ctx context.Context, filterID, dimensionN
 
 	// check the available filter dimensions.
 	// 1. A filter should always have a minimum of 2 dimensions
-	if len(filter.Dimensions) < 3 {
+	if len(filter.Dimensions) < 1 {
```
ChrysmOre, line 345: Should this have not gone down to 2 rather than 1? Also comment will need updating

saminahbab: done

j-s-rawat (approved):
LGTM

PR #155, opened 2022-11-09
fix: validation from 2 dimensions -> 1

saminahbab:
### What
Copy of validation trello ticket 5925
already approved, just sorting out concourse faults through another branch


thomasridd (approved):
LGT Jit on original branch - good enough for me

PR #159, opened 2022-11-16
Remove options href from API response

andre-urbani:
### What

Remove options href from API response
Trello - https://trello.com/c/Pr9kKvmV/5944-remove-options-href-from-api-response

### How to review

Confirm changes match requested changes in Trello ticket

### Who can review

Anyone

PR #165, opened 2022-11-24
Release/1.15.0

andre-urbani:
### What

[reinstate ID field](https://github.com/ONSdigital/dp-cantabular-filter-flex-api/commit/c1d675d24f4035407e337c60605880421980ce8d)

### How to review

Confirm there are no unexpected commits and tests pass

### Who can review

Anyone

### DIFF
--- a/features/get-dataset-json.feature
+++ b/features/get-dataset-json.feature
@@ -476,15 +476,12 @@ Feature: Get Dataset JSON
           "dimension_name":"city",
           "options":[
             {
-              "href":"http://hostname/code-lists/city/codes/0",
               "id":"0"
             },
             {
-              "href":"http://hostname/code-lists/city/codes/1",
               "id":"1"
             },
             {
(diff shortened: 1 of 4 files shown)


features/get-dataset-json.feature
```diff
@@ -476,15 +476,12 @@ Feature: Get Dataset JSON
           "dimension_name":"city",
           "options":[
             {
-              "href":"http://hostname/code-lists/city/codes/0",
               "id":"0"
```
ChrysmOre, line 479: How was this working before? Doesn't look like you changed this in the code :/

andre-urbani: It was removed in this PR as per the ticket - https://github.com/ONSdigital/dp-cantabular-filter-flex-api/pull/159 

but I removed the whole Options field rather than just the href by mistake, so ID needed to be reinstated

ChrysmOre (approved):

PR #171, opened 2022-12-07
New/default categorisation

saminahbab:
### What

Describe what you have changed and why.

### How to review

Describe the steps required to test the changes.

### Who can review


api/validation.go
```diff
@@ -74,11 +84,48 @@ func (api *API) validateAndHydrateDimensions(v dataset.Version, dims []model.Dim
...
+	}
+	return hydratedDimensions, nil
+}
+
 // getCantabularDimensions pulls full dimension information from Cantabular using the names of the provided
 // dimensions.
 // NOTE: when we hydrate the dimensions, we will be using the name as the id, and filling out the dimensions
 // using the same value for both.
-func (api *API) getCantabularDimensions(ctx context.Context, dimensions []model.Dimension, pType string) ([]model.Dimension, error) {
+func (api *API) getCantabularDimensions(ctx context.Context, dimensions []model.Dimension, pType string, postDimension bool) ([]model.Dimension, error) {
```
j-s-rawat, line 128: Do we need postDimension ?

api/validation.go
```diff
@@ -60,10 +60,20 @@ func (api *API) validateAndHydrateDimensions(v dataset.Version, dims []model.Dim
 	}
 
 	if v.IsBasedOn.Type == cantabularMultivariateTable {
-		resp, err := api.getCantabularDimensions(ctx, dims, pType)
+
+		if postDimension {
```
j-s-rawat, line 64: For a cantabularMultivariateTable if the portDimension is 'false' we are not following the process in which dimensions were populated in case of cantabularFlexibleTable https://github.com/ONSdigital/dp-cantabular-filter-flex-api/blob/develop/api/validation.go#L28

api/validation.go
saminahbab: yeah we do need it as the changes in this ticket are just for POST dimension, but also concern POST filter in multivariate cases

api/validation.go
saminahbab: yes we are I believe... as we then will be hitting this block `
		resp, err := api.getCantabularDimensions(ctx, dims, pType, postDimension)
		if err != nil {
			return nil, errors.Wrap(err, "failed to get dimensions from Cantabular")
		}

		return resp, nil`

api/validation.go
j-s-rawat: But where is is input param being used in that function?

api/validation.go
j-s-rawat: For filter-flex table type **getCantabularDimensions** is being called for the range of dims, but for multivariates if portDimension is false we are not doing the same. So either cantabularFlexibleTable is over-doing it or cantabularMultivariateTable is under-doing (or I didn't understood it well)
`for _, d := range dims {
			if d.IsAreaType != nil && *d.IsAreaType {
				areaCount++
				geodims, err := api.getCantabularDimensions(ctx, []model.Dimension{d}, pType, false)
				if err != nil {
					return nil, errors.Wrap(err, "failed to get geography dimension from Cantabular")
				}
				geodim = geodims[0]
			}
		}`

api/categorisation.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/ONSdigital/dp-api-clients-go/v2/cantabularmetadata"
+	"github.com/pkg/errors"
+)
+
+/*
+CheckDefaultCategorisation checks the default categorisation of a given dimension
+so that we store the correct parent dimension for a given set of options?
+*/
+func (api *API) CheckDefaultCategorisation(dimName string, datasetName string) (string, string, error) {
+
```
ChrysmOre, line 16: Unnecessary whitespace

api/categorisation.go
```diff
@@ -0,0 +1,68 @@
...
+	ctx := context.Background()
+	labelMap := make(map[string]string)
+	cats, err := api.ctblr.GetCategorisations(ctx, cantabular.GetCategorisationsRequest{
+		Dataset:  datasetName,
+		Variable: dimName,
+	})
+	if err != nil {
+		return "", "", err
```
ChrysmOre, line 24: Wrap error

api/categorisation.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	names := make([]string, 0)
+
+	if len(cats.Dataset.Variables.Edges) > 0 {
```
ChrysmOre, line 29: No need to check len(slice/map) before ranging. If len == 0 then block will be skipped

api/categorisation.go
```diff
@@ -0,0 +1,68 @@
...
+	if len(names) == 0 {
+		return "", "", errors.New("no categorisations recieved for variable")
+	}
+
+	defaultCat, err := api.metadata.GetDefaultClassification(ctx, cantabularmetadata.GetDefaultClassificationRequest{
+		Dataset:   datasetName,
+		Variables: names,
+	})
+	if err != nil {
+
```
ChrysmOre, line 62: Unnecessary whitespace, wrap error

api/validation.go
```diff
@@ -60,10 +60,20 @@ func (api *API) validateAndHydrateDimensions(v dataset.Version, dims []model.Dim
...
+			hydrated, err := api.HydrateMultivariateDimensions(dims, pType)
+			if err != nil {
+				return nil, err
```
ChrysmOre, line 67: wrap error

api/validation.go
```diff
@@ -60,10 +60,20 @@ func (api *API) validateAndHydrateDimensions(v dataset.Version, dims []model.Dim
...
+			}
+			return hydrated, nil
+		}
+
+		resp, err := api.getCantabularDimensions(ctx, dims, pType, postDimension)
```
ChrysmOre, line 72: This is bizzarre. You're passing the postDimension bool here but if postDimension == true it will never reach here. My instinct is that passing a 'PostDimension' bool around these validation functions isn't the correct approach. It's certainly not scalable (we can't just add another bool for each endpoint that needs something different done). Is there no way to rearrange it so the extra logic for PostDimension is handled in the PostDimensionHandler?

ChrysmOre: Seems like maybe after retrieving the response from ValidateAndHydrateDimensions you could then put it through the logic for replacing with default categorizations

api/categorisation.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	"context"
+
+	"github.com/ONSdigital/dp-api-clients-go/v2/cantabular"
+	"github.com/ONSdigital/dp-api-clients-go/v2/cantabularmetadata"
+	"github.com/pkg/errors"
+)
+
+/*
```
ChrysmOre, line 11: It's idiomatic to use // style comments rather than /* except when debugging/temporarily omitting code

api/categorisation.go
```diff
@@ -0,0 +1,60 @@
...
+		Dataset:  datasetName,
+		Variable: dimName,
+	})
+	if err != nil {
+		return "", "", errors.Wrap(err, "failed to check default categorisation")
+	}
+
+	names := make([]string, 0)
+
+	for _, edge := range cats.Dataset.Variables.Edges {
```
ChrysmOre, line 24: Unneccesary to be checking the length of these slices/maps before ranging over them

api/categorisation.go
```diff
@@ -0,0 +1,60 @@
...
+	defaultCat, err := api.metadata.GetDefaultClassification(ctx, cantabularmetadata.GetDefaultClassificationRequest{
+		Dataset:   datasetName,
+		Variables: names,
+	})
+	if err != nil {
+		return "", "", errors.Wrap(err, "failed to check default categorisation")
+	}
+
+	return defaultCat.Variable, labelMap[defaultCat.Variable], nil
+
```
ChrysmOre, line 59: Unnecessary whitespace

api/validation.go
```diff
@@ -60,10 +60,20 @@ func (api *API) validateAndHydrateDimensions(v dataset.Version, dims []model.Dim
 	}
 
 	if v.IsBasedOn.Type == cantabularMultivariateTable {
+
```
ChrysmOre, line 63: Unnecessary whiterspace

api/validation.go
```diff
@@ -74,6 +84,43 @@ func (api *API) validateAndHydrateDimensions(v dataset.Version, dims []model.Dim
...
+	for _, dim := range dimensions {
+		finalDimension := dim.Name
+		node, err := api.getCantabularDimension(ctx, pType, dim.Name)
+		if err != nil {
+			return nil, errors.Wrap(err, "error in cantabular response")
+		}
+
+		finalDimension, finalLabel, err = api.CheckDefaultCategorisation(node.Name, pType)
+		if err != nil {
+			return nil, err
```
ChrysmOre, line 103: wrap error?

api/validation.go
```diff
@@ -74,6 +84,43 @@ func (api *API) validateAndHydrateDimensions(v dataset.Version, dims []model.Dim
...
+		hydratedDimensions = append(hydratedDimensions, model.Dimension{
+			Name:                  finalDimension,
+			ID:                    finalDimension,
+			Label:                 finalLabel,
+			DefaultCategorisation: finalDimension,
+			Options:               dim.Options,
+			IsAreaType:            dim.IsAreaType,
+			FilterByParent:        dim.FilterByParent,
+		})
+
```
ChrysmOre, line 119: Unnecessary whitespace

api/categorisation.go
j-s-rawat: The current style is added to make it obvious that we are taking the values from either 'MapFrom' or from 'IsSourceOf' . The query returns both the sections in the result. If this is to be changed, then we will have to check the size of 'names' or 'labelMap' to determine if we have already populated the values and should not try the block starting from line 38.

api/validation.go
```diff
@@ -60,10 +60,20 @@ func (api *API) validateAndHydrateDimensions(v dataset.Version, dims []model.Dim
 	}
 
 	if v.IsBasedOn.Type == cantabularMultivariateTable {
```
j-s-rawat, line 62: Do we want to deal multivariate differently than flexible, won't adding a 'or' condition for multivariate here would address it https://github.com/ONSdigital/dp-cantabular-filter-flex-api/blob/cbc959241993cb2477272c32b0dc911aa5d76cd4/api/validation.go#L25

api/categorisation.go
```diff
@@ -0,0 +1,54 @@
...
+			}
+		} else if len(edge.Node.IsSourceOf.Edges) > 0 {
+			for _, sourceOf := range edge.Node.IsSourceOf.Edges {
+				names = append(names, sourceOf.Node.Name)
+				labelMap[sourceOf.Node.Name] = sourceOf.Node.Label
+			}
+		}
+
+	}
+	println(names[0])
```
ChrysmOre, line 40: debug

api/filter_dimensions.go
```diff
@@ -89,10 +89,23 @@ func (api *API) addFilterDimension(w http.ResponseWriter, r *http.Request) {
...
+	var dimensions []model.Dimension
+
+	if v.IsBasedOn.Type == cantabularFlexibleTable {
+		dimensions, err = api.validateAndHydrateDimensions(v, []model.Dimension{req.Dimension}, filter.PopulationType)
+		if err != nil {
+			api.respond.Error(ctx, w, statusCode(err), errors.Wrap(err, "failed to validate dimensions"))
+			return
+		}
+	} else if v.IsBasedOn.Type == cantabularMultivariateTable {
+
```
ChrysmOre, line 101: Extraneous whitespace

api/filter_dimensions.go
```diff
@@ -89,10 +89,23 @@ func (api *API) addFilterDimension(w http.ResponseWriter, r *http.Request) {
...
+		dimensions, err = api.HydrateMultivariateDimensions([]model.Dimension{req.Dimension}, filter.PopulationType)
+		if err != nil {
+			api.respond.Error(ctx, w, statusCode(err), errors.Wrap(err, "failed to validate dimensions"))
+			return
+		}
+	} else {
+		api.respond.Error(ctx, w, http.StatusInternalServerError, errors.New("unsupported filter type"))
```
ChrysmOre, line 108: Missed return

api/filter_dimensions.go
```diff
@@ -89,10 +89,23 @@ func (api *API) addFilterDimension(w http.ResponseWriter, r *http.Request) {
...
-		return
+	var dimensions []model.Dimension
+
+	if v.IsBasedOn.Type == cantabularFlexibleTable {
+		dimensions, err = api.validateAndHydrateDimensions(v, []model.Dimension{req.Dimension}, filter.PopulationType)
+		if err != nil {
+			api.respond.Error(ctx, w, statusCode(err), errors.Wrap(err, "failed to validate dimensions"))
+			return
+		}
+	} else if v.IsBasedOn.Type == cantabularMultivariateTable {
```
ChrysmOre, line 100: Won't this force paths not from POST /dimensions down this way when we don't want them to be? The existing checks for multivariates from other endpoints were correct no? This isn't the best way to be attempting to do this.

api/validation.go
```diff
@@ -74,6 +84,42 @@ func (api *API) validateAndHydrateDimensions(v dataset.Version, dims []model.Dim
 	}
 }
 
+// Used for POST Dimension when the filter type is multivariate.
+// Extra logic required to hydrate dimension using the default categorisation of the dimension.
+func (api *API) HydrateMultivariateDimensions(dimensions []model.Dimension, pType string) ([]model.Dimension, error) {
```
ChrysmOre, line 89: I don't like the naming and the comment.. HydrateMultiVariateDimensions suggests all multivariate hydration should be going through this but the comment (and my understanding) suggests is just the one off case for POST dimensions.

ChrysmOre (commented):
It looks like the case Jit mentioned of only validating dimension options if the dimension changes is not covered

saminahbab:
> It looks like the case Jit mentioned of only validating dimension options if the dimension changes is not covered
@ChrysmOre it is now covered, felt like the write thing to do was to set as default empty list in the case that the option changes to some default categorisation

j-s-rawat (approved):
LGTM

PR #176, opened 2023-01-10
use filterAPIURL and existing Version URL for new filter links

ChrysmOre:
### What

Update link generation in new filter to use FilterAPIURL and existing link in Version document

### How to review

Check changes make sense

### Who can review

Anyone


andre-urbani (approved):
LGTM

PR #177, opened 2023-01-10
patch: add default categorisation to response

saminahbab:
### What

Add default categorisation flag to filter flex api 
ammend tests to include default categorisation to all dimension responses

### DIFF
--- a/api/contract.go
+++ b/api/contract.go
@@ -203,7 +203,6 @@ func (d *dimensionItem) fromDimension(dim model.Dimension, host, filterID string
 	d.Label = dim.Label
 	d.FilterByParent = dim.FilterByParent
 	d.DefaultCategorisation = dim.DefaultCategorisation
-
 	d.Links = dimensionItemLinks{
 		Self: model.Link{
 			HREF: dimURL,
@@ -288,3 +287,4 @@ type DatasetJSONLinks struct {
 	Self            model.Link `json:"self"`
 	Version         model.Link `json:"version"`
 }
+
--- a/api/contract.go
+++ b/api/contract.go
@@ -186,11 +186,12 @@ type updateFilterDimensionResponse struct {
 }
 
 type dimensionItem struct {
-	ID             string             `json:"id"`
-	Name           string             `json:"name"`
-	Label          string             `json:"label"`
-	FilterByParent string             `json:"filter_by_parent,omitempty"`
-	Links          dimensionItemLinks `json:"links"`
+	ID                    string             `json:"id"`
+	Name                  string             `json:"name"`
+	Label                 string             `json:"label"`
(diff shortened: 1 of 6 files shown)


api/contract.go
```diff
@@ -201,6 +202,8 @@ func (d *dimensionItem) fromDimension(dim model.Dimension, host, filterID string
 	d.Name = dim.Name
 	d.Label = dim.Label
 	d.FilterByParent = dim.FilterByParent
+	d.DefaultCategorisation = dim.DefaultCategorisation
+
```
ChrysmOre, line 206: blank line?

features/multivariate_put_dimension.authorized.feature
```diff
@@ -183,4 +185,3 @@ Feature: Multivariate Feature Dimensions Private Endpoints
       "total_count": 2
     }
     """
```
ChrysmOre, line 187: MIssing CR

features/filter_dimension.authorised.feature
```diff
@@ -117,6 +117,7 @@ Feature: Filter Outputs Private Endpoints Enabled
           "name": "geography",
           "id": "city",
           "label": "City",
+          "default_categorisaion":  "",
```
ChrysmOre, line 120: Adding these blank values is no different from omitting completely. You should add a value if you want to capture any new functionality.

api/contract.go
```diff
@@ -186,11 +186,12 @@ type updateFilterDimensionResponse struct {
...
-	ID             string             `json:"id"`
-	Name           string             `json:"name"`
-	Label          string             `json:"label"`
-	FilterByParent string             `json:"filter_by_parent,omitempty"`
-	Links          dimensionItemLinks `json:"links"`
+	ID                    string             `json:"id"`
+	Name                  string             `json:"name"`
+	Label                 string             `json:"label"`
+	FilterByParent        string             `json:"filter_by_parent,omitempty"`
+	DefaultCategorisation string             `json:"default_categorisaion"`
```
ChrysmOre, line 193: Would we want to omitempty?

features/filter_dimension.authorised.feature
saminahbab: Nick mentioned that he needed a clear signal for the frontend controllers to understand when there is a default categorisation, so we thought of including it always, with the empty string representing a nil value, thus no default categorisation.

api/contract.go
saminahbab: i dont think so based on the conversation me and Jit had

features/multivariate_put_dimension.authorized.feature
saminahbab: should be fixed now? right side shows new lines added

j-s-rawat (approved):
LGTM

PR #186, opened 2023-02-10
Added support for custom dataset

j-s-rawat:
### What
Ticket : https://trello.com/c/ELlEBmEn/6022-create-end-point-for-new-custom-filter-flex-api

 Added functionality to create a custom filter based on a default dataset for a given population type.

### How to review

All test are :green_circle: 

### Who can review

Anyone


api/filters.go
```diff
@@ -197,6 +198,158 @@ func (api *API) createFilter(w http.ResponseWriter, r *http.Request) {
...
+			w,
+			statusCode(err),
+			Error{
+				err:     errors.Wrap(err, "failed to create filter"),
+				logData: logData,
+			},
+		)
+		return
+	}
+	// don't return dimensions in response
```
saminahbab, line 347: blank line between 346 and 347 would be more consistent if we intend to keep the comment

saminahbab (changes requested):

saminahbab (approved):

PR #188, opened 2023-02-14
Fix/master rebase

saminahbab:
merged master into develop to fix a conflict on one of our release branches

### DIFF
--- a/features/get-dataset-json.feature
+++ b/features/get-dataset-json.feature
@@ -1116,7 +1116,7 @@ Feature: Get Dataset JSON
     {
       "request": {
         "dimension":       "city",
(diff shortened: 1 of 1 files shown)


features/get-dataset-json.feature
```diff
@@ -1116,7 +1116,7 @@ Feature: Get Dataset JSON
     {
       "request": {
         "dimension":       "city",
-        "populationType":  "Example",
```
rafahop, line 1119: This change does not seem to be coming from `master` as this scenario does not exist in master. Not sure how it's ended up here if this PR is just to bring develop and master in line?

PR #190, opened 2023-02-16
upgrade dp-api-clients-go to 2.228.0

andre-urbani:
### What

Upgrade `dp-api-clients-go` version to `v2.228.0`

### How to review

Confirm we are using `v2.228.0` of `dp-api-clients-go`

### Who can review

Anyone


redhug1 (approved):
LGTM

PR #209, opened 2023-05-11
fix for quality statement/url

andre-urbani:
### What

Fix for QualityStatementText and QualityStatementURL not showing in filtered dimension response

Should be `node.QualityStatementText` and `node.QualitySummaryURL` instead of using `dim`, as `dim` is just the dimensions that we're passing into that function, rather than what's being returned from the `getCantabularDimension` function

### How to review

Changes make sense, tests pass

### Who can review

Anyone


andre-urbani:
Closed due to yesterday's Giithub issues not starting CI checks. Unable to start them manually

PR #221, opened 2024-01-09
Add temporary debugging code to dump http headers on specific requests

jessjenkins:
### What

Add temporary debugging code to dump http headers on specific requests

This will enable us to add a url parameter on specific requests and get a dump of the request's internal http request headers in the logs. For safety reasons, only the values of configured headers will be dumped out in order to prevent logging of sensitive values, eg. auth tokens. Also this functionality is only enabled when the debug key is set so causes no performance hit if unintentionally deployed to prod. 

This code is only intended to be deployed as far as staging so it can be tested with CloudFlare. 

To use, the env vars will be configured in vault…
- `ROUTER_DEBUG_KEY` = some value only know to us to stop anyone public using it (debugging disabled if blank)
- `ROUTER_DEBUG_HEADERS` = comma-separated list of headers whose values are alowed to be dumped to the logs (eg. `"X-Forwarded-For,Accept,Accept-Encoding"`

Requests can be made to trigger the debugging by adding the URL parameter `debug_key=some_value`.

NB. Also bumps adependency version that causes an audit failure

### How to review

Ensure the code does as described, especially that it will be disabled by default and only apply to a very limited and specific subset of requests.

Ensure tests pass.

### Who can review

Anyone but me.

### DIFF
--- a/api/dataset.go
+++ b/api/dataset.go
@@ -3,6 +3,7 @@ package api
 import (
 	"context"
 	"net/http"
+	"os"
 	"strings"
 
 	"github.com/ONSdigital/dp-api-clients-go/v2/cantabular"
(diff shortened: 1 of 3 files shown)


gedge (approved):
looks sane

PR #234, opened 2025-02-14
Feature/add url rewriting

andre-urbani:
### What

Added URL rewriting for the following GET endpoints - 

`/filters/{id}`
`/filters/{id}/dimensions`
`/filters/{id}/dimensions/{dimension}/options`
`/filter-outputs/{id}`
`/datasets/{dataset_id}/editions/{edition}/versions/{version}/json`
`/datasets/{dataset_id}/editions/{edition}/versions/{version}/census-observations`

Jira ticket - https://jira.ons.gov.uk/browse/DIS-2338

### How to review

Confirm that URL rewriting has been added on the above endpoints
Confirm that URL rewriting on these endpoints has been feature flagged
Changes make sense
Tests pass

### Who can review

Anyone


api/filter_outputs.go
```diff
@@ -32,6 +35,58 @@ func (api *API) getFilterOutput(w http.ResponseWriter, r *http.Request) {
...
+		if err != nil {
+			api.respond.Error(
+				ctx,
+				w,
+				statusCode(err),
+				Error{
+					err:     errors.Wrap(err, "failed to build self link"),
+					message: "failed to build self link",
+					logData: log.Data{
+						"id": fID,
```
aryan-p03, line 66: can we add the link which failed to rewrite here too?

api/filters.go
aryan-p03, line 1: Same issue with adding the link to the logData

config/config.go
aryan-p03, line 1: Is there a `config_test.go` file to add tests for these new configs?

features/filter_dimension-url-rewriting-enabled.feature
```diff
@@ -0,0 +1,168 @@
...
+        },
+        "published": true,
+        "population_type": "Example",
+        "type": "flexible"
+      }
+    ]
+    """
+
+  Scenario: Get filter dimensions successfully
+    And I set the "X-Forwarded-Proto" header to "https"
```
aryan-p03, line 111: All unit tests should still work without the `X-Forwarded-Proto` set (sorry it's a new dp-net update)

features/mock/generator.go
```diff
@@ -37,9 +38,18 @@ func (g *Generator) Timestamp() time.Time {
...
-	t, _ := time.Parse(time.RFC3339, TestTimestamp)
+	t, err := time.Parse(time.RFC3339, TestTimestamp)
+	if err != nil {
+		panic(fmt.Sprintf("invalid timestamp format: %v", err))
+	}
+
+	seconds := t.Unix()
+	if seconds < 0 || seconds > math.MaxUint32 {
+		panic(fmt.Sprintf("timestamp %d out of uint32 range", seconds))
+	}
```
aryan-p03, line 49: I'm not sure panic() is the best thing to do here

generator/generator.go
```diff
@@ -38,8 +39,13 @@ func (g *Generator) UUID() (uuid.UUID, error) {
 // UniqueTimestamp generates a timestamp of the current time in the
 // special format required by mongoDB
 func (g *Generator) UniqueTimestamp() primitive.Timestamp {
+	now := time.Now().Unix()
+	if now < 0 || now > math.MaxUint32 {
+		panic(fmt.Sprintf("timestamp %d out of uint32 range", now))
+	}
```
aryan-p03, line 45: Same comment about the panic(), I'm assuming the linter brought this up

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/ONSdigital/dp-cantabular-filter-flex-api
 
-go 1.22
+go 1.23.0
```
aryan-p03, line 3: can leave it as 1.23

go.mod
```diff
@@ -15,21 +17,24 @@ replace github.com/gogo/protobuf => github.com/gogo/protobuf v1.3.2
 // [CVE-2021-3121] CWE-129: Improper Validation of Array Index
 replace google.golang.org/protobuf => google.golang.org/protobuf v1.33.0
 
+// [CVE-2024-45339] CWE-377: Insecure Temporary File
+replace github.com/golang/glog => github.com/golang/glog v1.2.4
+
 // to fix: [CVE-2023-32731] CWE-Other
 replace google.golang.org/grpc => google.golang.org/grpc v1.64.0
```
aryan-p03, line 24: Best to avoid having any `replace` statements if possible

aryan-p03 (changes requested):

config/config.go
```diff
@@ -40,6 +41,7 @@ type Config struct {
 	MaxRowsReturned              int           `envconfig:"MAX_ROWS_RETURNED"`
 	Mongo                        mongo.MongoDriverConfig
 	KafkaConfig                  KafkaConfig
+	EnableURLRewriting           bool
```
franmoore05, line 44: This should have an associated envconfig value like the others so it can be set in dp-configs

api/contract.go
```diff
@@ -225,15 +230,42 @@ type dimensionItem struct {
...
+		return
+	}
 	filterURL := fmt.Sprintf("%s/filters/%s", host, filterID)
 	dimURL := fmt.Sprintf("%s/dimensions/%s", filterURL, dim.Name)
 
+	parsedURL, err := url.Parse(cfg.CantabularFilterFlexAPIURL)
+	if err != nil {
+		log.Println("Error parsing URL:", err)
+		return
+	}
```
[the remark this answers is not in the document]
andre-urbani: I think it has to be parsed in this case because we're using the `dimensionItem` receiver for this function, meaning we're not using the `API` receiver as we are in most of the other functions (which is how we are able to reference the api.cfg value directly)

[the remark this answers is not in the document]
aryan-p03: Within the function `getFilterDimension` the api is referenced here. We can add the parsed one into the api and then pass this value into `fromDimension` by adding a new parameter.

config/config.go
andre-urbani: Added `CantabularFilterFlexAPIURL` to the config test

api/contract.go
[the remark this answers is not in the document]
andre-urbani: I don't think this is possible because it creates a large chain of function signatures that need to be updated, eventually leading to multiple `api.Router.Get()` functions which expect a `Get(pattern string, h http.HandlerFunc)` signature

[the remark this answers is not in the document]
andre-urbani: Actually I guess I can just use the receiver instead of changing the function signature

features/mock/generator.go
```diff
@@ -37,9 +40,18 @@ func (g *Generator) Timestamp() time.Time {
...
+
+	seconds := t.Unix()
+	if seconds < 0 || seconds > math.MaxUint32 {
+		log.Fatal(context.Background(), fmt.Sprintf("timestamp %d out of uint32 range", seconds))
+	}
+
 	return primitive.Timestamp{
-		T: uint32(t.Unix()),
+		T: uint32(seconds),
```
aryan-p03, line 54: I think we can change this back to how it was originally and just ignore it in the `golangci.yml` file with a comment saying it does not need to be changed since unit32 can store dates until February 7, 2106

generator/generator.go
```diff
@@ -38,8 +41,14 @@ func (g *Generator) UUID() (uuid.UUID, error) {
...
 func (g *Generator) UniqueTimestamp() primitive.Timestamp {
+	now := time.Now().Unix()
+	if now < 0 || now > math.MaxUint32 {
+		log.Fatal(context.Background(), fmt.Sprintf("timestamp %d out of uint32 range", now))
+		return primitive.Timestamp{T: 0, I: 0}
+	}
+
 	return primitive.Timestamp{
-		T: uint32(time.Now().Unix()),
+		T: uint32(now),
```
aryan-p03, line 51: same as other timestamp comment

PR #239, opened 2025-02-25
add missing header for /filters/{id}/dimensions/{dimension}

andre-urbani:
### What

Add missing header for `/filters/{id}/dimensions/{dimension}`

As done for the following endpoints here -
https://github.com/ONSdigital/dp-cantabular-filter-flex-api/blob/develop/api/filters.go#L466
https://github.com/ONSdigital/dp-cantabular-filter-flex-api/blob/develop/api/filter_dimensions.go#L394
https://github.com/ONSdigital/dp-cantabular-filter-flex-api/blob/develop/api/filter_dimension_options.go#L291
https://github.com/ONSdigital/dp-cantabular-filter-flex-api/blob/develop/api/filter_outputs.go#L17 

### How to review

Confirm the above has been done, tests pass

### Who can review

Anyone

### DIFF
--- a/api/filter_dimensions.go
+++ b/api/filter_dimensions.go
@@ -317,6 +317,8 @@ func (api *API) getFilterDimension(w http.ResponseWriter, r *http.Request) {
 	fID := chi.URLParam(r, "id")
 	dim := chi.URLParam(r, "dimension")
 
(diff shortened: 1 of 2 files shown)


aryan-p03 (changes requested):
Is a component test needed as well for this endpoint?

aryan-p03 (approved):

